PR #17, opened 2015-04-28
5836 remote api server errors made obvious

mzats:


sdk/go/arvadosclient/arvadosclient.go
```diff
@@ -20,13 +20,31 @@ import (
...
+	HttpStatusCode    int
+	HttpStatusMessage string
+
+	// Additional error details from response body.
+	ErrorDetails []string
 }
 
-func (e ArvadosApiError) Error() string { return e.error.Error() }
+func (e RemoteApiServerError) Error() string {
+	s := fmt.Sprintf("API server (%s) returned %d: %s.",
```
tomclegg: Having all the information is certainly good but I feel like this formatting obscures the actual problem/error by putting it after a bunch of details.

    API server (zzzzz.arvadosapi.com:443) returned 422: Unprocessable Entity. Additional details: #<NoMemoryError: failed to allocate memory>

If we have ErrorDetails, the server has gone to the trouble of giving us a human-readable error string, so I think we should feature this first, and append the technical details -- starting with the more likely clues -- in case that message isn't enough.

    API error: #<NoMemoryError: failed to allocate memory> (422 Unprocessable Entity from zzzzz.arvadosapi.com:443)

Or, if we didn't get an error message in JSON:

    API error: 500 Internal Server Error from zzzzz.arvadosapi.com:443

Suggest dropping trailing period, cf. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

sdk/go/arvadosclient/arvadosclient.go
```diff
@@ -149,30 +167,36 @@ func (this ArvadosClient) CallRaw(method string, resource string, uuid string, a
 	}
 
 	defer resp.Body.Close()
-	errorText := fmt.Sprintf("API response: %s", resp.Status)
+	return nil, NewRemoteApiServerError(this.ApiServer, resp)
+}
+
+func NewRemoteApiServerError(ServerAddress string, resp *http.Response) RemoteApiServerError {
```
tomclegg: Suggest `newRemoteAPIServerError`. (Or is there a reason to export this?)

(Acronyms are conventionally [API, not Api](https://github.com/golang/go/wiki/CodeReviewComments#initialisms))

Seems to me `RemoteAPIServer` is either redundant or possibly incorrect depending on how you define "remote", and naming the type `APIError` would be fine -- this version will look like `arvadosclient.RemoteAPIServerError` to the caller which seems to belabor the point and even incorrectly suggest a 500-like "server error". This type covers errors like "permission denied", which I would describe as an API or transaction error but not a server error.

sdk/go/arvadosclient/arvadosclient.go
mzats: Thanks! I switched to the following two flavors:

With details:

    arvados API server error: #<NoMemoryError: failed to allocate memory> (422: Unprocessable Entity) returned by zzzzz.arvadosapi.com:443

Without details:

    arvados API server error: 422: Unprocessable Entity returned by zzzzz.arvadosapi.com:443

sdk/go/arvadosclient/arvadosclient.go
mzats: Makes sense. I changed the names to `APIServerError` and `newAPIServerError()`. I left the `Server` string in the error name, because I worry that users will misread `APIError` as `SDKError` and incorrectly conclude (as I did) that this error occurred locally on their machine.

PR #19, opened 2015-07-31
Adds useful log warnings re: repo failures

jrandall:
Instruments the find_commit_range function such that it logs an
explanatory warning in any case when it fails to find any refs.

When no refs are found, find_commit_range returns an empty list
which typically causes its caller to throw an exception. Without
these added warnings, that exception is not at all helpful in
tracking down the underlying issue.


tomclegg:
This is great, thanks!

PR #21, opened 2015-09-03
Adds error handling for command not found

jrandall:
Raises a "command not found" error if arv-* commands are not
found.


jrandall:
This has been incorporated into PR #22

PR #22, opened 2015-09-04
Implements `arv get` subcommand

jrandall:
Implements `arv view` feature as in https://arvados.org/issues/7204

Refactors portions of `arv_edit` into separate functions so `arv_view` can share them.


sdk/cli/bin/arv
```diff
@@ -85,7 +85,15 @@ def init_config
 end
 
 
-subcommands = %w(copy create edit keep pipeline run tag ws)
+subcommands = %w(copy create edit get keep pipeline run tag ws)
+
+def exec_bin bin, opts
+  @bin_path = `which #{bin}`.strip
```
tomclegg: `@bin_path` should be just a local variable `bin_path` here. (aiui, `@bin_path` is an instance variable of "main", aka a global, and although practically speaking it looks harmless, globals do have a way of becoming harmful over time, so probably best to fix it now.)

sdk/cli/bin/arv
```diff
@@ -150,14 +160,7 @@ end
...
 
-  content = case global_opts[:format]
-            when 'json'
-              Oj.dump(initial_obj, :indent => 1)
-            when 'yaml'
-              initial_obj.to_yaml
-            else
-              abort "Unrecognized format #{global_opts[:format]}"
-            end
+  content = get_obj_content initial_obj, global_opts
```
tomclegg: :+1: moving this out to a method, but now that it's further away from its counterpart in "Load the new object" below, perhaps it's a good time to add an "else abort" in the "Load" section too, so if the format lists go out of sync we throw a nice loud exception right away instead of having a mysterious nil floating about.

sdk/cli/bin/arv
```diff
@@ -294,15 +284,50 @@ def arv_edit client, arvados, global_opts, remaining_opts
...
+    puts "open an interactive text editor on a text representation (json or\n"
+    puts "yaml, use --format) and then update the object.  Will use 'nano'\n"
+    puts "by default, customize with the EDITOR or VISUAL environment variable.\n"
+    exit 255
+  end
+
+  if not $stdout.tty?
+    puts "Not connected to a TTY, cannot run interactive editor."
+    exit 1
+  end
```
tomclegg: (make sure this doesn't get added back accidentally after disappearing in #23)

tomclegg:
I think this deserves a couple of tests. The testing infrastructure for "arv" is somewhat lacking, but not a total loss -- should be possible to adapt some of the code in `sdk/cli/test/test_arv-get.rb` et al.

jrandall:
Still sorting out my dev/test environment so I can write some tests for this...

colin-nolan:
Tests for `arv get` are ready for review.

sdk/cli/test/test_arv-get.rb
```diff
@@ -1,251 +1,192 @@
...
+  INVALID_FIELD_NAME = "invalid"
 
-  def test_file_to_file_no_overwrite_file_in_dir
-    File.open './tmp/foo', 'wb' do |f|
-      f.write 'baz'
-    end
+  # Tests that a valid Arvados object can be retrieved in a supported format
+  # using: `arv get [uuid]`. Given all other `arv foo` commands return JSON
+  # when no format is specified, JSON should be expected in this case.
+  def test_get_valid_object_no_format_specified()
```
tomclegg: Nit: Most of this branch is closer to https://github.com/bbatsov/ruby-style-guide than prevailing style (which is surely a move in the right direction) but it does say to [omit the parens when no params](https://github.com/bbatsov/ruby-style-guide#method-parens)

sdk/cli/test/test_arv-get.rb
```diff
@@ -1,251 +1,192 @@
...
-    end
-    out, err = capture_subprocess_io do
-      assert_arv_get false, @@foo_manifest_locator + '/foo', 'tmp/foo'
-    end
-    assert_match /Local file tmp\/foo already exists/, err
-    assert_equal '', out
-    assert_equal 'baz', IO.read('tmp/foo')
-  end
+  # UUID for an Arvados object that does not exist
+  NON_EXISTANT_OBJECT_UUID = "qr1hi-tpzed-p9yk1lihjsgwew0"
```
tomclegg: existant &rarr; existent

use "zzzzz" instead of "qr1hi" to avoid conflating "uuid does not exist" and "uuid is not in our namespace and does not exist" (currently those cases are handled identically, so I'm just thinking of future federation features and human readers)

colin-nolan:
Made suggested changes.

tomclegg:
@colin-nolan thanks! and sorry for the delay in merging.

PR #23, opened 2015-09-04
Fixes 7205 (https://arvados.org/issues/7205)

jrandall:
Removes unnecessary check that stdout is a TTY and moves printing of `arv edit` messages out-of-band

Messages regarding update of object after edit are now printed out-of-band to STDERR rather than to STDOUT where they might mix with the output of non-interactive "editors".


tomclegg:
Agree 100% with both changes, thank you!

PR #24, opened 2015-09-10
7262: blacklists pbr 1.7.0

jrandall:


jrandall:
pbr 1.8.0 just came out on PyPI and it also has the same issue as 1.7.0. I've blacklisted pbr>=1.7.0 for now.

PR #25, opened 2015-09-11
7307: Fixes mocked up CGI output and adds comment regarding error

jrandall:
Fixes #7307

### DIFF
--- a/services/arv-git-httpd/git_handler_test.go
+++ b/services/arv-git-httpd/git_handler_test.go
@@ -25,7 +25,7 @@ func (s *GitHandlerSuite) TestEnvVars(c *check.C) {
 	}
 	h := newGitHandler()
 	h.(*gitHandler).Path = "/bin/sh"
-	h.(*gitHandler).Args = []string{"-c", "echo HTTP/1.1 200 OK; echo Content-Type: text/plain; echo; env"}
+	h.(*gitHandler).Args = []string{"-c", "printf 'Content-Type: text/plain\r\n\r\n'; env"}
 	os.Setenv("GITOLITE_HTTP_HOME", "/test/ghh")
 	os.Setenv("GL_BYPASS_ACCESS_CHECKS", "yesplease")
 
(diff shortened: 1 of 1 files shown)


services/arv-git-httpd/git_handler_test.go
```diff
@@ -41,13 +41,14 @@ func (s *GitHandlerSuite) TestEnvVars(c *check.C) {
 }
 
 func (s *GitHandlerSuite) TestCGIError(c *check.C) {
+	log.Printf("git_handler_test: TestCGIError() - expecting a SplitHostPort error, so it is safe to ignore!")
```
tomclegg: How about we just rename the test to TestCGIErrorOnSplitHostPortError and/or change "bogus" to "intentionally.broken.host.with.no.port"? I think we might go crazy if we start trying to explain away error messages that get logged during tests by adding more log messages to the tests...

services/arv-git-httpd/git_handler_test.go
```diff
@@ -25,7 +25,7 @@ func (s *GitHandlerSuite) TestEnvVars(c *check.C) {
 	}
 	h := newGitHandler()
 	h.(*gitHandler).Path = "/bin/sh"
-	h.(*gitHandler).Args = []string{"-c", "echo HTTP/1.1 200 OK; echo Content-Type: text/plain; echo; env"}
+	h.(*gitHandler).Args = []string{"-c", "printf 'Content-Type: text/plain\r\n\r\n'; env"}
```
tomclegg: :+1:, thanks.

services/arv-git-httpd/git_handler_test.go
jrandall: I was emulating other behaviour I observed when running the tests. While trying to get the dev/test environment working, I very much appreciated tests that told me to expect a failure/warning message. I think the particular one I am thinking of is in the install phase regarding it being safe to ignore package missing warnings.

brettcs: That's fair—but the reason we write out a special message in that case is because we have no control over the installer code.  I think our own test code follows the pattern Tom describes.

jrandall: Ok, happy with changing the name of the test and the bogus value to make it clear that a failure is expected.

PR #26, opened 2015-09-11
7308: adds print logging to arv-git-httpd tests

jrandall:
Fixes #7308


tomclegg:
I'm not keen on adding boilerplate to tests. Can we get the same benefit by making it easier to turn on (or find out about) "go test -gocheck.vv"?

jrandall:
If some level of gocheck verbosity were turned on by default in run-tests, I think that would do it, yes!

jrandall:
https://github.com/curoverse/arvados-dev/pull/3

PR #27, opened 2015-09-11
7310: adds git v2 style regex to match http 500 error

jrandall:


tomclegg:
Tested successfully on wheezy (git 1.7) and jessie (git 2.1.4), thanks!

PR #28, opened 2015-09-14
7324: fixes /dev/stderr issue by creating and catting a fifo 

jrandall:
Fixes 7324 by adding {{ACCESSLOG}} to nginx.conf template and populating it with the path to a fifo created by run_test_server.py in TEST_TMPDIR. Immediately after starting nginx, another process is started to cat the access logs to stderr. 

This isn't the most elegant solution, but I'm not sure how else to solve this given the inability to configure nginx to write its access_logs directly to stdout/stderr.

### DIFF
--- a/sdk/python/tests/run_test_server.py
+++ b/sdk/python/tests/run_test_server.py
@@ -441,6 +441,7 @@ def run_nginx():
     nginxconf['GITSSLPORT'] = find_available_port()
     nginxconf['SSLCERT'] = os.path.join(SERVICES_SRC_DIR, 'api', 'tmp', 'self-signed.pem')
     nginxconf['SSLKEY'] = os.path.join(SERVICES_SRC_DIR, 'api', 'tmp', 'self-signed.key')
+    nginxconf['ACCESSLOG'] = os.path.join(TEST_TMPDIR, 'nginx_access_log.fifo')
 
     conftemplatefile = os.path.join(MY_DIRNAME, 'nginx.conf')
     conffile = os.path.join(TEST_TMPDIR, 'nginx.conf')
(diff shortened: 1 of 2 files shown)


sdk/python/tests/run_test_server.py
```diff
@@ -452,12 +453,16 @@ def run_nginx():
 
     env = os.environ.copy()
     env['PATH'] = env['PATH']+':/sbin:/usr/sbin:/usr/local/sbin'
+    os.mkfifo(nginxconf['ACCESSLOG'], 0700)
```
brettcs: Because `TEST_TMPDIR` is static, this needs to deal with the fact that the FIFO might already exist.  Otherwise this happens on subsequent test runs:

```
  File "./run_test_server.py", line 628, in <module>
    run_nginx()
  File "./run_test_server.py", line 456, in run_nginx
    os.mkfifo(nginxconf['ACCESSLOG'], 0700)
OSError: [Errno 17] File exists
rake aborted!
Command failed (pid 7530 exit 1): [["python", "./run_test_server.py", "start_nginx"]]
/home/brett/repos/arvados/apps/workbench/test/test_helper.rb:165:in `block in check_output'
…
```

Everything else looks good to me.

jrandall: Good catch! I didn't notice because I'm testing in docker containers that have been built from scratch each time. I think the right way to fix this is probably to clean up by removing the ACCESSLOG fifo after the test. I thought about just checking if the fifo exists and only making it if necessary, but that could lead to trouble if a test is interrupted in such a way that the cat doesn't get killed (and in that case, it should be an error, but it can print a more informative message).

brettcs: Unfortunately, you can lose both ways, right?  If you lose such that cat never dies (e.g., because nginx fails to start), then reusing an existing FIFO is a bad idea.  But if you lose such that cat does die but the FIFO isn't removed (e.g., the system shuts down between when the tests call `run_nginx` and when they call `stop_nginx`), reporting that as an error is not so helpful.

Normally everything in TEST_TMPDIR we rewrite with impunity.  Since you don't have that option, I think you want to avoid it completely and instead use mktmpfifo.  Or, since that doesn't exist, mkdtemp and put the FIFO in there.  `stop_nginx` can still clean up as you describe, but we'll also recover well from more dramatic failures.

brettcs:
Thanks very much for this.  I'm merging this version, with just a small tweak to make sure we only ignore OSErrors that are raised because the FIFO doesn't exist.

PR #30, opened 2015-10-18
Adds preliminar support for Arvados docker running on OSX

brainstorm:
The check still needs to be made more robust (FIXME), but at least it does instantiate/run the docker containers in a boot2docker and virtualbox backed docker setup on OSX Yosemite.

/ping @guillermo-carrasco, might interest you too.


brainstorm:
Superseeded by arvbox.

PR #32, opened 2015-10-28
"websocket_address" is now "websockets_address", update it in install doc.

aflyhorse:
As title.


tomclegg:
The config really is called `websocket_address` (although admittedly `websockets_address` might have been a better choice). I'm guessing you were misled by the appearance of `websockets_address` in `services/api/config/application.yml.example`. I've just fixed that to match the rest of the world in a5d766f.

aflyhorse:
Nice, but why websockets_address still works in my instance :( . Maybe we need some syntax check in the config yml, to remind user of any invalid parameters.
Thanks for your explanation.

PR #33, opened 2015-11-02
Add hint values and others into config examples, as well as minor doc fixes

aflyhorse:
Add hint values into api and workbench's config yml examples. This make it clear what should be insert there, and reduced the users' effort to type in manually.
Also add some more options, which might interest the users, into them.


aflyhorse:
appended some fix on installation docs.

apps/workbench/config/application.yml.example
```diff
@@ -24,6 +24,9 @@ production:
...
 
   # You probably also want to point to your API server.
-  arvados_login_base: https://arvados.local:3030/login
-  arvados_v1_base: https://arvados.local:3030/arvados/v1
+  arvados_login_base: https://uuid_prefix.your.domain/login
+  arvados_v1_base: https://uuid_prefix.your.domain/arvados/v1
   arvados_insecure_https: false
+
+  # You might also want to name your workbench
+  site_name: ~
```
tomclegg: What's the rationale for adding this? `site_name` already has a reasonable default. (This file should be the smallest set of configs that can get Workbench started -- it doesn't need to mirror `application.default.yml`.)

services/api/config/application.yml.example
```diff
@@ -18,25 +18,35 @@ production:
...
+  sso_provider_url: https://uuid_prefix.your.domain
+  workbench_address: https://workbench.uuid_prefix.your.domain
+  websocket_address: wss://ws.uuid_prefix.your.domain/websocket
   #git_repositories_dir: ~
   #git_internal_dir: ~
+  #git_repo_ssh_base: "git@git.uuid_prefix.your.domain:"
+  #git_repo_http_base: https://git.uuid_prefix.your.domain/
+  #auto_admin_first_user: false
+  #crunch_job_wrapper: :slurm_immediate
+  #sso_insecure: false
```
tomclegg: The real defaults (and docs) for these configs are already in `application.default.yml` -- we should avoid having an independent list of defaults here. (I'm not sure what those existing `git_*` comments are doing there but presumably we should remove them.)

services/api/config/application.yml.example
```diff
@@ -18,25 +18,35 @@ production:
   secret_token: ~
   blob_signing_key: ~
   sso_app_secret: ~
-  sso_app_id: ~
-  sso_provider_url: ~
-  workbench_address: ~
-  websocket_address: ~
+  sso_app_id: arvados-server
+  sso_provider_url: https://uuid_prefix.your.domain
```
tomclegg: Should be `auth.your.domain` to match the install docs (except the parts of the install docs that say `sso.your-site.com`, which should also be fixed, whoops)

services/api/config/application.yml.example
aflyhorse: The installation document said the user need to add these lines manually. So either the user don't need to add them (and the cluster can work fine without them), or the user need to add them himself. If the user *does* need to add them manually, why not leave a hint there, ready to be replaced or uncomment?
```auto_admin_first_user``` is an interesting feature and should be noted in the install doc. It can replace the procedure of adding the first user, make it admin and activate it. This saves a lot of time.
```sso_insecure``` should also be noted here and in doc, in case the user is using self-signed certs in test environment.

File: services/api/config/application.yml.example (hunk 3 above)
Comment (aflyhorse) [inline services/api/config/application.yml.example]: this was my fault.

File: apps/workbench/config/application.yml.example (hunk 1 above)
Comment (aflyhorse) [inline apps/workbench/config/application.yml.example]: IMHO, the production section of this file should contain the essential options, as well as the ones which the user is mostly interested in. Also, anything noted in the install documentation should appear here with or without comment. So the user can be sure he is changing the right file and the right options, without fearing he just added another unrecognizable row.

Comment (tomclegg) [inline apps/workbench/config/application.yml.example]: I agree it's more convenient at setup time to have everything right there in the editable config file. The problem is that your customized copy of the example file isn't touched during upgrades, so after a few upgrades the convenience bites back: the comments (including "default is X") no longer match what the software actually does. "Look in the actual defaults/docs file for more info" is less convenient during the initial setup, but at least it doesn't start misleading you later.

`site_name` isn't essential so it doesn't belong in the minimal file. "Mostly interesting" varies from one user to another. How about improving the comments in the example file to encourage people to look in `application.default.yml` and decide which options are most interesting to them?

We could also point out in the example file that if you prefer a big editable template (and don't mind the comments/defaults going out of sync when you upgrade the software) you can simply use a copy of `application.default.yml` as the starting point for your `application.yml`. AFAIK that should work fine.

File: services/api/config/application.yml.example (hunk 2 above)
Comment (tomclegg) [inline services/api/config/application.yml.example]: [Install docs](http://doc.arvados.org/install/install-api-server.html#omniauth) do say you need to add these *if* you aren't using the default/recommended locations for some reason. Normally, the defaults are fine and you don't need to do anything here.

Mentioning `auto_admin_first_user` in the install docs seems like a good idea -- it's surely more convenient if your API server isn't publicly accessible. The method given in the install doc (using rails console) is safer in that you get a chance to confirm you're giving root to the correct account before you actually do it (even though the docs don't show the part of the transcript where rails console displays the user details).

I do like adding `sso_insecure` here, right beside the other `sso_*` configs.

Comment (aflyhorse) [inline services/api/config/application.yml.example]: The lines I added was from the [Git](http://doc.arvados.org/install/install-arv-git-httpd.html) and [Crunch](http://doc.arvados.org/install/install-crunch-dispatch.html) install doc. They both said "add the line".
I arranged the order of options in the same way as ```application.default.yml```. I've reordered them on your request.

File: apps/workbench/config/application.yml.example (hunk 1 above)
Comment (aflyhorse) [inline apps/workbench/config/application.yml.example]: I would simply leave the hint as "option_name: ~" so it won't effect your decision, unless the option name itself changed. However I understand your concerning so I withdrew my changes on this.
Maybe we could add the following line at the end of every install doc
> For more advanced configs, please refer to ```application.default.yml``` in ```/var/www/arvados-xxx/config```.

**Comment by aflyhorse** [comment]
In crunch log:
```
salloc: Granted job allocation 10
Can't locate Net/SSL.pm in @INC (you may need to install the Net::SSL module) (@INC contains: /etc/perl /usr
BEGIN failed--compilation aborted at /usr/share/perl5/Arvados.pm line 62.
Compilation failed in require at /var/www/arvados-api/shared/vendor_bundle/ruby/2.1.0/gems/arvados-cli-0.1.2
BEGIN failed--compilation aborted at /var/www/arvados-api/shared/vendor_bundle/ruby/2.1.0/gems/arvados-cli-0
salloc: Relinquishing job allocation 10
```
Add ```libcrypt-ssleay-perl``` to resolve it. This is the Debian equivalent of CentOS ```perl-Crypt-SSLeay```.

aflyhorse:
Rebased to the current master branch.

tetron:
Outdated

PR #36, opened 2016-02-04
8345 updates llfuse requirement and API to >=0.42

jrandall:


tfmorris:
Thanks for the contribution. I know there have been a number of offline conversations about llfuse, but just to update any followers, we upgraded to llfuse to v1.2 in July 2017. We currently have the version pinned to <1.3.4 due to an installer issue with that version, so 1.3.3 is the latest supported version.

Closing this as outdated.

PR #38, opened 2016-02-18
8484: adds sanity check on number of collections retrieved

jrandall:
Changes GetCollections() to return an error if the number
of collections retrieved is less than the initial number
of collections.

### DIFF
--- a/services/datamanager/collection/collection.go
+++ b/services/datamanager/collection/collection.go
@@ -227,6 +227,17 @@ func GetCollections(params GetCollectionsParams) (results ReadCollections, err e
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


tomclegg:
I like the premise, I just wonder if we can do something a bit stronger and with less risk of false positives. "Deleting some really old collections" might be the time when you least want datamanager to skip garbage collection, as well as the most likely way to generate false positives here.

How about:

After retrieving everything, do one more API request with filter `modified_at<=latestModificationDate` and limit 0. We don't need to see the items, just check `ItemsAvailable<=len(results.UUIDToCollection)`.

jrandall:
@tomclegg: I have implemented the improved sanity check as you suggested.

PR #39, opened 2016-02-18
8485: changes GetCollection loop to more reliably fetch all collections

jrandall:
Modifies the loop termination condition in GetCollection so that
it continues until there are no more items available according to
the API server (returned in collections.ItemsAvailable).

Modifies the query code so that it uses an offset to page through
results in case an entire batch has equal modified_at timestamps.

Adds count of collections remaining to the log output printed for 
each batch.

### DIFF
--- a/services/datamanager/collection/collection.go
+++ b/services/datamanager/collection/collection.go
@@ -131,7 +131,8 @@ func GetCollections(params GetCollectionsParams) (results ReadCollections, err e
 	sdkParams := arvadosclient.Dict{
 		"select":  fieldsWanted,
 		"order":   []string{"modified_at ASC"},
-		"filters": [][]string{[]string{"modified_at", ">=", "1900-01-01T00:00:00Z"}}}
+		"filters": [][]string{[]string{"modified_at", ">=", "1900-01-01T00:00:00Z"}},
+		"offset": 0}
 
 	if params.BatchSize > 0 {
 		sdkParams["limit"] = params.BatchSize
@@ -176,9 +177,10 @@ func GetCollections(params GetCollectionsParams) (results ReadCollections, err e
 
 	// These values are just for getting the loop to run the first time,
 	// afterwards they'll be set to real values.
-	previousTotalCollections := -1
-	totalCollections := 0
-	for totalCollections > previousTotalCollections {
+	remainingCollections := 1
+	var totalCollections int
+	var previousTotalCollections int
+	for remainingCollections > 0 {
 		// We're still finding new collections
 
 		// Write the heap profile for examining memory usage
@@ -194,6 +196,12 @@ func GetCollections(params GetCollectionsParams) (results ReadCollections, err e
 			return
 		}
 
+		// Update count of remaining collections
+		remainingCollections = collections.ItemsAvailable - params.BatchSize - sdkParams["offset"].(int)
+		if remainingCollections < 0 {
+			remainingCollections = 0
+		}
+
 		// Process collection and update our date filter.
 		latestModificationDate, maxManifestSize, totalManifestSize, err := ProcessCollections(params.Logger,
 			collections.Items,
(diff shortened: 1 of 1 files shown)


services/datamanager/collection/collection.go
```diff
@@ -131,7 +131,8 @@ func GetCollections(params GetCollectionsParams) (results ReadCollections, err e
 	sdkParams := arvadosclient.Dict{
 		"select":  fieldsWanted,
 		"order":   []string{"modified_at ASC"},
```
tomclegg: Even though the API should do it for us automatically, for clarity (and a safety belt) I think we should add "uuid asc" here explicitly -- we're depending on the server to give us a consistent order (among items with identical modified_at) from one page to the next.

services/datamanager/collection/collection.go
```diff
@@ -194,6 +196,12 @@ func GetCollections(params GetCollectionsParams) (results ReadCollections, err e
 			return
 		}
 
+		// Update count of remaining collections
+		remainingCollections = collections.ItemsAvailable - params.BatchSize - sdkParams["offset"].(int)
```
tomclegg: Use `len(collections.Items)`, not `params.BatchSize` -- asking for `limit=100` doesn't guarantee we get 100 (even if we're not on  the last page).

services/datamanager/collection/collection.go
```diff
@@ -202,16 +210,23 @@ func GetCollections(params GetCollectionsParams) (results ReadCollections, err e
...
 
 		// update counts
 		previousTotalCollections = totalCollections
 		totalCollections = len(results.UUIDToCollection)
 
 		log.Printf("%d collections read, %d new in last batch, "+
+			"%d remaining, "+
 			"%s latest modified date, %.0f %d %d avg,max,total manifest size",
 			totalCollections,
 			totalCollections-previousTotalCollections,
```
tomclegg: This can also be just `len(collections.Items)` ...and then we don't need the `previousTotalCollections` var at all.

services/datamanager/collection/collection.go
```diff
@@ -202,16 +210,23 @@ func GetCollections(params GetCollectionsParams) (results ReadCollections, err e
 		if err != nil {
 			return results, err
 		}
-		sdkParams["filters"].([][]string)[0][2] = latestModificationDate.Format(time.RFC3339)
+		if sdkParams["filters"].([][]string)[0][2] != latestModificationDate.Format(time.RFC3339) {
+			sdkParams["filters"].([][]string)[0][2] = latestModificationDate.Format(time.RFC3339)
+			sdkParams["offset"] = 0
+		} else {
+			sdkParams["offset"] = sdkParams["offset"].(int) + params.BatchSize
```
tomclegg: `len(collections.Items)`

services/datamanager/collection/collection.go
jrandall: Being explicit is always good. Adding this.

services/datamanager/collection/collection.go
jrandall: Ok, I'll set `batchCollections := len(collections.Items)` once and use that throughout.

services/datamanager/collection/collection.go
jrandall: I think you are actually referring to `%d new in last batch` here, but that is not the same thing as `len(collections.Items)`. I've modified this to make it more explicit:

`%d (%d new) in last batch`

services/datamanager/collection/collection.go
jrandall: I've made this change (using the aforementioned `batchCollections` rather than calling `len` multiple times), but for safety it also necessitates an additional sanity check that we did in fact receive some items in the batch. Otherwise, it would be theoretically possible to loop forever if the query returns no items (and no explicit error) but has non-zero ItemsAvailable. 

Maybe that is "not possible" given the current API server, but I think it is prudent to avoid a potential infinite loop and it doesn't cost much. 

Making the suggested change and adding the extra sanity check.

tomclegg:
Thanks @jrandall

PR #41, opened 2016-02-19
8497: makes datamanager `-collection-batch-size` an option and sets default to 1000

jrandall:
Raises the BatchSize for fetching collections from 50 to 1000 to improve performance.


jrandall:
Batch size now configurable at run-time using `-collection-batch-size` option (default: 1000)

PR #44, opened 2016-02-29
7263: even better busy behavior (also including fix for 8575) 

jrandall:
Uses `IO::Select` in `readfrompipes` to select on a list of fds rather than slowly looping through them all.
Reaps as many children in `reapchildren` as are ready to be reaped, rather than one at a time.
Queries the API server for new tasks in `reapchildren` in a batch rather than one task at a time.


jrandall:
Now rebased onto master for ease of merging

tomclegg:
Github didn't follow all the rebasing and cherry-picking but all of these changes are merged in fbd5446 &rarr; 9ddb8e3 (first two commits), and ec51cd7 (third commit). Thanks @jrandall!

PR #45, opened 2016-05-18
9232: Adds histogram of replication level distribution

jrandall:


jrandall:
This was implemented in keep-balance

PR #47, opened 2016-05-19
9154: adds final pass through collection retrieval loop

jrandall:
does a final pass with timestamp == final timestamp to ensure that we have all entries up to the final timestamp (otherwise the subsequent items_available check could fail).


jrandall:
data-manager has been superceded by keep-balance

PR #48, opened 2016-07-13
Patch socket configuration for Apple systems

adambom:
This was causing failures when using the `arv` CLI on OSX.

### DIFF
--- a/sdk/python/arvados/keep.py
+++ b/sdk/python/arvados/keep.py
@@ -9,6 +9,7 @@ import Queue
 import re
 import socket
 import ssl
+import sys
 import threading
 import timer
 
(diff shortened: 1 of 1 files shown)


sdk/python/arvados/keep.py
```diff
@@ -22,6 +22,16 @@
...
 
 
+# Monkey patch TCP constants when not available (apple). Values sourced from:
+# http://www.opensource.apple.com/source/xnu/xnu-2422.115.4/bsd/netinet/tcp.h
+if not hasattr(socket, 'TCP_KEEPALIVE'):
+    socket.TCP_KEEPALIVE = 0x010
+if not hasattr(socket, 'TCP_KEEPINTVL'):
+    socket.TCP_KEEPINTVL = 0x101
+if not hasattr(socket, 'TCP_KEEPCNT'):
+    socket.TCP_KEEPCNT = 0x102
```
tomclegg: I don't think we can assume OSX is the only platform where those constants aren't already set. How about putting this block in an `if sys.platform == 'darwin':`?

adambom: I can add that, but keep will still fail to open socket connections regardless of platform if these constants are not defined. I think this would fail on Windows as well.

tomclegg: Do those Apple values succeed in Windows? I don't think we can expect the Apple values to work on all platforms where Python doesn't define them for us. E.g., `TCP_KEEPINTVL` is 0x5 in linux, 0x200 in FreeBSD... Setting the wrong sockopts seems like a recipe for confusion, at best.

tomclegg:
@adambom thanks!

PR #49, opened 2016-07-18
Fix path to docker metadata file

adambom:
Docker images do not have a `json` file at their root, but rather, a set of
layers, which use a `json` file to manage metadata for the layer. At the root,
the image has a file for metadata named `<hash>.json`. This change instructs
keep to look at that file for the metadata rather than the `json` file.

An issue for this bug has been created here: https://dev.arvados.org/issues/9628


tfmorris:
Thanks for your contribution. Sorry we didn't give you more timely feedback on it. #50 provided a similar fix which was merged and now that Arvados fully supports Docker 1.10, support is built into the core. Customers upgrading from Docker 1.9 to 1.10+ can find upgrade instructions here: http://doc.arvados.org/install/migrate-docker19.html

Closing this as outdated.

PR #52, opened 2016-11-12
Fixes API server to return default columns for collections

jrandall:
Prior to this fix, if a request has a query param `select=`
(as opposed to just omitting `select` entirely), the API
server would return a full set of columns rather than
enforcing its own defaults.

Fixes #10517, assuming this was a bug and not a feature.


tomclegg:
Thanks @jrandall. Merged in 0b5d04b along with some tests (but I cherry-picked as b865434 instead of merging your branch, so github didn't notice the connection...oops)

PR #53, opened 2016-11-12
9998 add count option to list method

jrandall:
When `count` (default: `true`) is false, the API server will not query 
the database for the count of items and it will not return `items_available` 
in the JSON output.

This is expected to offer a major performance improvement when
tables are large and the count is not needed, since postgres
can otherwise take far longer to answer the COUNT() query than
to answer the main query and return the requested list.

I had originally called this option `no_count` but that conflicted with 
the `arv` cli's automatic handling (in Trollop) of negated flag values. 
With the present implementation, all of the following would continue 
to include `items_available` in the output:
`arv collection list`
`arv collection list -c`
`arv collection list --count`

Whereas the negated count param would exclude it:
`arv collection list --no-count`

Also adds the `Count` (and missing `Distinct`) params to the Go SDK's ResourceListParams.

Finally, changes keep-balance to use `count=false` for all the collection retrievals except for at the beginning and end when a count is needed.

### DIFF
--- a/doc/sdk/cli/reference.html.textile.liquid
+++ b/doc/sdk/cli/reference.html.textile.liquid
@@ -60,7 +60,7 @@ Arguments accepted by the @list@ subcommand include:
     --order, -r <s>:     Order in which to return matching users.
    --select, -s <s>:     Select which fields to return
      --distinct, -d:     Return each distinct object.
-     --no-count, -n:     Don't count items_available.
+         --no-count:     Don't count items_available.
 </pre>
 
 The @--filters@ option takes a string describing a JSON list of filters on which the returned resources should be returned. Each filter is a three-element list of _[field, operator, value]_, where the _operator_ may be one of @=@, @<@, @<=@, @>@, @>=@, @!=@, @like@, or @ilike@.
--- a/doc/sdk/cli/reference.html.textile.liquid
+++ b/doc/sdk/cli/reference.html.textile.liquid
@@ -59,7 +59,8 @@ Arguments accepted by the @list@ subcommand include:
   --filters, -f <s>:     Conditions for filtering users.
     --order, -r <s>:     Order in which to return matching users.
    --select, -s <s>:     Select which fields to return
-     --distinct, -d:     Return each distinct object
+     --distinct, -d:     Return each distinct object.
+     --no-count, -n:     Don't count items_available.
 </pre>
 
 The @--filters@ option takes a string describing a JSON list of filters on which the returned resources should be returned. Each filter is a three-element list of _[field, operator, value]_, where the _operator_ may be one of @=@, @<@, @<=@, @>@, @>=@, @!=@, @like@, or @ilike@.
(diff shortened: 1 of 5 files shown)


tomclegg:
Currently keep-balance relies on `ItemsAvailable` to differentiate between "Received an entire page with the same modified_at timestamp" (bug, cannot proceed) and "no more collections left to retrieve". I think this change will make it always conclude the latter, which could result in data loss if the former is true.

Perhaps when we reach the "was it the last page?" decision for the first time we should just set `params.Count = true` and proceed as usual -- then, the next iteration will either get an empty page (if we're done) or give us a non-zero `ItemsAvailable` so the existing "identical timestamps" check can be effective.

jrandall:
@tomclegg ok, I see the problem. I think the solution should be to just implement the same iteration that we had in datamanager (which can page through sets of collections with identical timestamps using offset). I know now that we are using nanosecond timestamps it should never happen in the wild, but given the importance of not missing collections, it doesn't hurt to implement it just in case something sadistic happens to the database and we wind up with a full page of collections with the same timestamp.

jrandall:
@tomclegg or I guess implementing the TODO would be even better (using a `uuid>y` filter to page instead of offset). The `if page.ItemsAvailable > len(page.Items)` comparison is not necessary as long as the next iteration of the loop can handle an empty page. The logic there would be, if the page is empty, we are only done if the `uuid>y` filter is not set. If a `uuid` filter is set, unset it and proceed to the next iteration. Does that make sense to you?

tfmorris:
It's a nit, but I have a small preference for -count={exact|none} default:exact to leave the door open for some type of approximate counting strategy.

tomclegg:
@jrandall yes, correctly handling the identical-timestamp cases sounds even better than reliably detecting it and bailing out.

The strategy sounds right, but I think we also need to change the timestamp filter back and forth between `=` and `>=`: otherwise records with `uuid<lastUUID && timestamp>identicalTimestamp` can be skipped over on the last page of identical-timestamp results.

jrandall:
@tfmorris ok, I do agree that -count={exact|none} is preferable. I'll make those changes when I get a chance. Postgres does have a simple way of getting an approximate row count for the entire table very quickly, and a somewhat more involved way of getting an approximate count for more sophisticated queries (https://wiki.postgresql.org/wiki/Count_estimate), which should still be faster than a full table scan. 

I won't implement anything like -count=estimate as part of this PR - but I'll create an issue to keep it on record as future functionality.

jrandall:
@tfmorris yes, I'd use the `uuid` filter in the same way we had used the `offset` in datamanager.

The procedure would be: if we detect a full page with identical timestamps, we temporarily add in the `uuid` filter and continue paging through using `uuid` (i.e. updating it to be greater than the last retrieved uuid) until we encounter a different timestamp, at which point we drop the `uuid` filter and revert back to only filtering on timestamp.

tomclegg:
Finally merged this, with exact/none and the keep-balance paging fix -- thanks @jrandall!

PR #54, opened 2016-11-15
update signature of SdNotify for go-systemd v14

jrandall:
should fix 10534 - will run local tests now.


tfmorris:
Thanks for the fix!

PR #55, opened 2016-11-15
10535 sdk python test callback race

jrandall:
fixes 10535

### DIFF
--- a/sdk/python/tests/test_events.py
+++ b/sdk/python/tests/test_events.py
@@ -335,7 +335,10 @@ class PollClientTestCase(unittest.TestCase):
         self.build_client(poll_time=.01)
         with self.callback_cond:
             self.client.start()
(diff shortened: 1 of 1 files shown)


jrandall:
Change 10.0s to a constant

tomclegg:
I changed the notification mechanism here from `threading.Condition()` to `threading.Event()` in a983bbe, which (according to my reasoning, and a few trials) fixes the race and eliminates the need for `sleep()`. Still added the 10-second timeouts proposed here though, so if the code breaks then the tests fail instead of hanging. @jrandall if a983bbe (in master) works for you too, I think we can close this PR. Either way, thanks for pinning this down!

tfmorris:
@jrandall I'm assuming that this fixed things, so closing the PR, but if not, please feel free to reopen.

PR #56, opened 2016-11-16
9897: Logs errors encountered by block prefetch workers.

colin-nolan:


tfmorris:
Ugh!

             except Exception:		              except Exception:
                  pass

Now there's an anti-pattern if there ever was one.

sdk/python/arvados/arvfile.py
```diff
@@ -516,7 +517,7 @@ def _block_prefetch_worker(self):
                     return
                 self._keep.get(b)
             except Exception:
-                pass
+                _logger.error(traceback.format_exc())
```
colin-nolan: I agree that knowing that a failure has occurred is a step in the right direction. However, that solution doesn't provide any insight into the nature and origin of the related fault. In addition, it obfuscates the location of that exception handler, making it difficult for someone not familiar with the codebase to know where to start.

A traceback provides much more useful information, including the location you suggest would be useful (albeit expressed without knowledge of the higher level context).

colin-nolan: I believe the purpose of this Pokémon exception handler is to suppress failures from unknown rare errors. Therefore, the message should scarcely be seen by users of the SDK. This being the case, it seems reasonable for it to be intended for developers.

With regards to the extra import: the import statement could be moved in the `except` block?

jrandall: apparently `_logger.exception()` does include a stack trace in the output

jrandall: Maybe include "Warning:" in the exception message as well so that people know this is not an expected exception.

colin-nolan: I don't believe that this is the case:
```
>>> _logger = logging.getLogger("test")
>>> def test():
...     _logger.exception("Some message")
... 
>>> test()
ERROR:test:Some message
None
```

tomclegg: @colin-nolan "`_logger.exception()` does include a stack trace" is true when called from an exception handler (as in ["This method should only be called from an exception handler."](https://docs.python.org/2/library/logging.html#logging.Logger.exception) :)

```python
import logging
logging.basicConfig()
_logger = logging.getLogger()
try:
    nan = 1/0
except:
    _logger.exception("boom")
```

```
ERROR:root:boom
  File "boom.py", line 5, in <module>
    nan = 1/0
ZeroDivisionError: integer division or modulo by zero
```

colin-nolan: Apologies; I stand corrected. I wasn't aware that this method's behaviour was context sensitive.

tomclegg:
@colin-nolan thanks!

PR #58, opened 2017-04-12
Make explicit support for Python 2 and 3

cclauss:
Arvados-DCO-1.1-Signed-off-by: Chris Clauss <cclauss@bluewin.ch>

### DIFF
--- a/sdk/python/setup.py
+++ b/sdk/python/setup.py
@@ -57,4 +57,8 @@ setup(name='arvados-python-client',
       tests_require=['pbr<1.7.0', 'mock>=1.0', 'PyYAML'],
       zip_safe=False,
       cmdclass={'egg_info': tagger},
+      classifiers=[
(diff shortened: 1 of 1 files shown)


tfmorris:
LGTM. Sorry for the delay in reviewing.

@tomclegg WDYT?

tomclegg:
LGTM.

Sorry this is red tape, but we now require a DCO sign-off line in each commit message. @cclauss, if you reword your commit to include this, I can merge. See https://dev.arvados.org/projects/arvados/wiki/Developer_Certificate_Of_Origin

Thanks!

cclauss:
@tomclegg Thoughts?

tfmorris:
@cclauss I just noticed that you added your DCO sign-off line to the Github comment which is probably why @tomclegg didn't notice it.

If you do a git commit --amend and add it to the end of commit message itself, we should be all set.

Sorry for all the hoops!

PR #59, opened 2017-10-08
fix s3_volume so that it can create empty recent/* objects on radosgw

jrandall:
Ensures that keepstore includes a `Content-Length: 0` header when creating empty objects. 

Fixes issue 12416.


tomclegg:
This looks like it should have been fixed in bd96be714d with a goamz fork (we ran into the same trouble with AWS when we started building with Go 1.8). Is it possible the problem still exists only when you build locally without using govendor? Or are you seeing it with our released packages, too?

jrandall:
Had this problem with released keepstore package (0.1.20170929182415.9e58fc0-1).

tomclegg:
Merged with a small refactoring in a3851eec6 to ensure the PutReader() case gets fixed too.

PR #60, opened 2017-12-20
Feature/compute driver openstack

lukaszimmermann:
This PR adds the OpenStack Compute Node driver for the Arvados Node Manager.

Some important remarks:
- the cloud names are assigned sequentially from compute-0 to compute-N; name is skipped if it is already 
  listed by `list_nodes()`. Hence, to avoid duplicate cloud names, a file lock is used which ensures that
  the code of `create_node()` is only processed once at a time, the filelock module must be installed in 
  Python.
- The node_secret is exported to the new node via cloud-init, which must therefore be installed on the
  used compute image. 
- The config must specify the location of the lockfile for creating new nodes.
- The config file must specify these IDs: `'image_id', 'flavor_id', 'network_id', 'security_group_id'`. Currently, only one item is allowed for each type.
- `sync_node` currently does not do anything useful. 
- the meta values associated with a compute node are set in the config.ini. Only nodes having all (!)
  meta values set to the chosen value will be listed in `list_nodes()`.

The logic of giving cloud names is different than in other drivers (does not use `compute-UUID`). I chose this method to avoid the need to rename the cloud node after it has been created. In `compute-N`, N is not the slot number (as one would want to have). This is due to the fact that the slot number has not been decided upon at the time the cloud node is created. It is selected once the first ping is received.

### DIFF
--- a/services/nodemanager/arvnodeman/computenode/driver/openstack.py
+++ b/services/nodemanager/arvnodeman/computenode/driver/openstack.py
@@ -1,0 +1,166 @@
+#!/usr/bin/env python
+# Copyright (C) The Arvados Authors. All rights reserved.
+#
+# SPDX-License-Identifier: AGPL-3.0
+
+from __future__ import absolute_import, print_function
+
+import functools
+import json
+import time
+import datetime
+import filelock 
+
+
+import libcloud.compute.providers as cloud_provider
+import libcloud.compute.types as cloud_types
+
+from . import BaseComputeNodeDriver
+from .. import arvados_node_fqdn, arvados_timestamp, ARVADOS_TIMEFMT
+
+cloud_init_config = """
+#cloud-config
+write_files:
+-   content: {}
+    owner: root:root
+    path: /var/tmp/arv-node-data/arv-ping-url
+    permissions: '0400'
+"""
+LOCKFILE_NAME = 'lockfile'
+
+class ComputeNodeDriver(BaseComputeNodeDriver):
+
+    DEFAULT_DRIVER = cloud_provider.get_driver(cloud_types.Provider.OPENSTACK)
+    SEARCH_CACHE = {}
+    
+    """
+    Compute node driver for OpenStack. It heavily uses libcloud drivers.
+    This class must implement the methods:
+
+    arvados_create_kwargs, sync_node, node_fqdn, node_start_time,
+    create_cloud_name
+    """
+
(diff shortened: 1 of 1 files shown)


tetron:
Thanks for you contribution!  We'll need some time to do a code review.  Would you be able to add some testing?  Also, does this depend on #61?

lukaszimmermann:
Yes, I can add tests soon.
This PR does not depend on #61, though I did not test so far with  `assign_node_hostname: true`

tfmorris:
From a process point of view, we need the commits signed off for substantial contributions like this.

Please review the requirements at https://dev.arvados.org/projects/arvados/wiki/Developer_Certificate_Of_Origin
and amend your commits to add Arvados-DCO-Signoff lines if you agree and meet all the requirements.

tetron:
Another question, is this something you're using now, i.e. you already have a functioning Arvados cluster on OpenStack, or is this an ongoing development?

lukaszimmermann:
I currently work on deploying Arvados on the de.NBI cloud. This is not a production system yet, but we have a CWL workflow that we want to offer soon. As de.NBI cloud uses OpenStack, I needed to write the compute node driver. I tested this the driver "a bit" and it seems to work fine. (Nodes are booted and shut down by the NodeManager as expected). The only problem that I see is that the `crunch_worker_state` is consistently `down` and that the nodes are all `unpaired`. I have not yet figured out why the pairing between Arvados nodes and compute nodes does not work yet.

tetron:
@lukaszimmermann There are a few additional steps required for the compute node to self-register after it is booted by node manager.  They are covered in http://doc.arvados.org/install/crunch2-slurm/install-compute-node.html and http://doc.arvados.org/install/install-compute-ping.html

If you have more questions about configuring your compute node image, consider asking the mailing list:

http://lists.arvados.org/mailman/listinfo/arvados

lukaszimmermann:
Thanks for the reference. I thoroughly read the install documentation and ended up using the same ping script that was described in the doc. The only difference is that I manage the dynamic DNS server externally (because I use `bind9` rather than `unbound`), so the ping script is extended with code that registers the newly created node at the DNS server.

I also saw that the `slurm_state` key never appears  in the node info section.  This results in the node reported as down. I still investigate why this is the case.

lukaszimmermann:
I added the Arvados-DCO-Signoff lines

tetron:
Closing old PRs

PR #61, opened 2017-12-20
Node must assign hostname via first ping if not set by Arvados

lukaszimmermann:
I experienced the issue that node `hostname`s   (from the Node model class) remain unassigned if 
`assign_node_hostname: false`
in the `application.yml`. 

I changed the logic such that the hostname must be included in the ping from the cloud node if
`assign_node_hostname: false`. The sent value will then be applied to the node model.


tetron:
Closing old PRs

PR #63, opened 2018-02-16
add ReserveExtraRAM config option to crunch-dispatch-slurm

jrandall:
fixes 13093


tetron:
Can you amend this with an arvados-DCO line?

https://dev.arvados.org/projects/arvados/wiki/Developer_Certificate_Of_Origin

PR #65, opened 2018-04-20
Fix LocalJumpError in arvados-login-sync

strangepleasures:


services/login-sync/bin/arvados-login-sync
```diff
@@ -57,7 +57,9 @@ begin
 
   pwnam = Hash.new()
   logins.reject! do |l|
-    return false if pwnam[l[:username]]
+    if pwnam[l[:username]]
+      false
+    end
```
tetron: This isn't right.  The remainder of the block needs to be inside an `else` otherwise the `if` statement evaluates false but then it proceeds to evaluate the rest of the block instead of rejecting the user as intended.

strangepleasures:
That's correct. I've made the required changes.

services/login-sync/bin/arvados-login-sync
```diff
@@ -57,18 +57,21 @@ begin
...
-    begin
-      pwnam[l[:username]] = Etc.getpwnam(l[:username])
-    rescue
-      if skip_missing_users
-        STDERR.puts "Account #{l[:username]} not found. Skipping"
-        true
-      end
+    if pwnam[l[:username]]
+      false
     else
```
tetron: Actually this can be simplified further to `if not pwnam[l[:username]]` and pull up the else clause.  The works because if statements without an "else" clause evaluate to nil, which means the login is not rejected (it is accepted.)

strangepleasures: Done

PR #67, opened 2018-05-24
add --submit-runner-vcpus argument to arvados-cwl-runner

jrandall:


mr-c:
This seems like a nice fix for complicated workflows (or at least those with big scatters, I'm guessing). Is it still relevant?

PR #69, opened 2018-06-01
add req_id as a tag to Rails.logger

jrandall:


tfmorris:
Thanks for your contribution!

PR #72, opened 2018-07-13
packages should not depend on setuptools

graingert:


graingert:
You already know that setuptools is installed here: https://github.com/curoverse/arvados/blob/master/sdk/python/setup.py#L11

tetron:
Hi @graingert 

Is this causing an actual problem for you or just a linting / style question?  I believe the reason for this dependency is that:

* We use `pkg_resources` at runtime (which is part of setuptools)
* We build operating system packages (deb and rpm) which *do not* run setup.py at install time
* The OS packages need an explicit dependency on setuptools in order to import `pkg_resources`

That said, the setuptools dependency could be moved to the OS package build if it causes a concrete problem for pip install.

graingert:
This causes a problem because we use `pip-compile` with `--generate-hashes` when we install the requirements.txt `setuptools` is considered a package without hashes defined

tetron:
Thanks.  I'll see what we can do.

PR #74, opened 2018-07-13
Allow hardcoding of protocol and port components of root_url

jrandall:
Fixes 13809 with the option to hardcode all relevant components of root_url in the arvados `application.yml` configuration.


tfmorris:
@jrandall Thanks for your patch!

For reference, the bug that this fixes is https://dev.arvados.org/issues/13809

PR #75, opened 2018-08-07
Allow Arvados model audit logging to be disabled

colin-nolan:
As discussed with @tomclegg, the number of audit logs created in our system overwhelms the database with logs and causes things to slow down dramatically.

This PR alllows audit logging to be disabled when the max audit log age is 0 and the audit log delete batch size is more than 0.


tomclegg:
Thanks @colin-nolan - sorry about the wait.

PR #78, opened 2018-08-20
14075 fix slow queries in python commands sdk `uploadfiles`

jrandall:
change `uploadfiles` to include pdh in name of collection rather then using `ensure_unique_names=True`
also don't bother attempting to list existing empty collections as this always exists

### DIFF
--- a/sdk/python/arvados/commands/run.py
+++ b/sdk/python/arvados/commands/run.py
@@ -1,4 +1,17 @@
 # Copyright (C) The Arvados Authors. All rights reserved.
+# Copyright (C) 2018 Genome Research Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


jrandall:
This is built against 1.1.4.20180607205857 as that is what we are running for our arvados-jobs containers currently and we don't want to make too many changes in the middle of an analysis workflow. I'll work on rebasing it to the current master once we are back up and running.

tetron:
I'm working on fixing this up and will merge it.

tfmorris:
Thanks @tetron -- and thanks @jrandall for the fix! 

To make it easier to find the original issue (because we don't use the Github issue tracker) here's a direct link https://dev.arvados.org/issues/14075

tetron:
This was merged in 5b8dfb5d5f4b34fa730e69c45c517f0cfb3fdc9f (github isn't picking it up because it was also rebased)

PR #80, opened 2018-08-23
limit concurrent slurm commands

jrandall:
use a semaphore channel to limit concurrent sbatch/scancel/scontrol commands to 3.
squeue is already limited to one at a time.

fixes 14110


tfmorris:
Thanks for your contribution!

Fixes https://dev.arvados.org/issues/14110

PR #81, opened 2018-09-02
add /keep empty collection mount to a-c-r

jrandall:


tetron:
Whats the reasoning for this?

jrandall:
This is a signal to crunch-run that it is ok to bind mount over `/keep`
(and therefore it is ok to implement the mounting of all subdirectories of
`/keep` within a single docker bind mount). This mechanism was suggested by
Tom Clegg as a simpler alternative to what I proposed.

This is a part of the fix to 13636 - the other part (of our implementation,
which is working) is a revised crunch-run which uses the `bindmapfuse` (
https://github.com/wtsi-hgi/bindmapfuse) fuse filesystem to provide limited
access to the underlying `arv-mount` filesystem. I.e. we use the
`bindmapfuse` configuration rather than docker bind mounts to limit which
portions of `/keep` are accessible to the container.

The reason for the empty mount at `/keep` is that without that, it wouldn't
necessarily be safe for crunch-run to decide on its own to bind mount
`/keep` (i.e. because perhaps the container has some other contents under
that directory that it needs to be able to see in addition to the
subdirectories that are bind mounted in).

tetron:
So this only makes sense with corresponding crunch-run changes?  I'm very confused by this, I think I understand the problem you're trying to solve but it seems like a very roundabout solution.

PR #85, opened 2019-02-14
Python 3: only import subprocess32 on py27

chapmanb:
Avoids the need to install and import subprocess32 when working
with any recent Python 3.x version. This helps with bioconda
install since subprocess32 is not built for py3.

### DIFF
--- a/sdk/cwl/arvados_cwl/runner.py
+++ b/sdk/cwl/arvados_cwl/runner.py
@@ -7,12 +7,16 @@ standard_library.install_aliases()
 from future.utils import  viewvalues, viewitems
 
 import os
+import sys
 import urllib.parse
 from functools import partial
 import logging
 import json
-import subprocess32 as subprocess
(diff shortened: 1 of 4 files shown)


ericBiagiotti:
Brad, We'll handle my suggestion on our end. Thanks again!

ericBiagiotti:
Hi Brad, can you amend your commit message to include the DCO sign off, i.e Arvados-DCO-1.1-Signed-off-by: Brad Chapman <email>

tfmorris:
Here's more detail on what you're certifying with the Developer Certificate of Origin: https://dev.arvados.org/projects/arvados/wiki/Developer_Certificate_Of_Origin

This is a new requirement that we added in the last year or two.

chapmanb:
Thanks, done. I'm not very good at paperwork so appreciate the heads up. Thanks again for including this.

PR #103, opened 2019-11-08
Optimize OkHttpClient resource usage

evgeny-sabirov:
Redesign OkHttpClientFactory:
- Singleton factory implementation
- Predefined default OkHttpClient instances
- Build OkHttpClient instances with shared factories/connection pools to optimize resource usage

Arvados-DCO-1.1-Signed-off-by: Evgeny Sabirov <evgeny.sabirov@genestack.com>

### DIFF
--- a/sdk/java-v2/src/main/java/org/arvados/client/api/client/BaseApiClient.java
+++ b/sdk/java-v2/src/main/java/org/arvados/client/api/client/BaseApiClient.java
@@ -34,9 +34,7 @@ abstract class BaseApiClient {
 
     BaseApiClient(ConfigProvider config) {
         this.config = config;
-        client = OkHttpClientFactory.builder()
-                .build()
-                .create(config.isApiHostInsecure());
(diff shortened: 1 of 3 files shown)


evgeny-sabirov:
As per the doc: https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/ _OkHttpClient_ instances should be shared. Right now when creating the single _ArvadosFacade_ instance **5** brand new OkHttpClient (!!!) client instances get instantiated: this is terrible to say the least. Now imagine users that use multiple _ArvadosFacade_ instances...
Furthermore, additional  _OkHttpClient_ client configuration can also be externalized with the existing _ConfigProvider_ mechanism and _OkHttpClientFactory_ can be adapted to consume the config with `.create(ConfigProvider config) `

evgeny-sabirov:
There's more: `ArvadosFacade.downloadFilesFromCollection(String collectionUuid, String pathToDownloadFolder)` actually creates new OkHttpClient instances per each downloaded chunk of data! 
See for yourself: 

alt text

@tetron i suggest to resovle this ASAP.

PR #107, opened 2019-12-03
Fix file token error

MajewskiKrzysztof:
In some cases `filePosition` is greater than `Integer.MAX_VALUE`

### DIFF
--- a/sdk/java-v2/src/main/java/org/arvados/client/logic/collection/FileToken.java
+++ b/sdk/java-v2/src/main/java/org/arvados/client/logic/collection/FileToken.java
@@ -12,7 +12,7 @@ import org.arvados.client.common.Characters;
 
 public class FileToken {
 
-    private int filePosition;
+    private long filePosition;
     private long fileSize;
     private String fileName;
     private String path;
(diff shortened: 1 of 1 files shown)


sdk/java-v2/src/main/java/org/arvados/client/logic/collection/FileToken.java
```diff
@@ -42,7 +42,7 @@ public String getFullPath() {
         return Strings.isNullOrEmpty(path) ? fileName : path + fileName;
     }
 
-    public int getFilePosition() {
+    public long getFilePosition() {
```
tfmorris: This is a breaking change to the public API, so we may need to introduce a new method instead.

Having said that, any code using the existing method and storing the result in an int is arguably broken and needs to be updated, so perhaps we'd be doing people a favor by making that obvious.

If we go that route, we'll need an update to the readme/release notes with a description of the incompatible API change.

tetron: The one situation I could see this causing a problem would be if someone was assigning the result of this to an `int`, which would begin to produce a warning or error if the return type is changed to `long`.  However, the original code is clearly wrong, so I don't think backwards compatibility over such a small change is a sufficient argument to preserve an incorrect method signature.

tetron:
Arvados ticket here https://dev.arvados.org/issues/15919

PR #129, opened 2020-05-20
util.py: python-api https certificate openssl override as is used in …

pjotrp:
…GNU Guix - feel free to override the commit message.

### DIFF
--- a/sdk/python/arvados/keep.py
+++ b/sdk/python/arvados/keep.py
@@ -375,6 +375,8 @@ class KeepClient(object):
                     curl.setopt(pycurl.HEADERFUNCTION, self._headerfunction)
                     if self.insecure:
                         curl.setopt(pycurl.SSL_VERIFYPEER, 0)
+                    else:
+                        curl.setopt(pycurl.CAPATH,os.path.dirname(arvados.util.ca_certs_path()))
                     if method == "HEAD":
                         curl.setopt(pycurl.NOBODY, True)
                     self._setcurltimeouts(curl, timeout, method=="HEAD")
@@ -473,6 +475,8 @@ class KeepClient(object):
                     curl.setopt(pycurl.HEADERFUNCTION, self._headerfunction)
                     if self.insecure:
                         curl.setopt(pycurl.SSL_VERIFYPEER, 0)
+                    else:
+                        curl.setopt(pycurl.CAPATH,os.path.dirname(arvados.util.ca_certs_path()))
                     self._setcurltimeouts(curl, timeout)
                     try:
                         curl.perform()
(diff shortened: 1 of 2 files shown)


tetron:
Could you amend with a developer certificate of origin (DCO) as described here https://github.com/arvados/arvados/blob/master/CONTRIBUTING.md

tetron:
(From chat) I think you found a bug, we're not using the same certificates when reading from keep.  Take a look at this code block for get()

https://github.com/arvados/arvados/blob/master/sdk/python/arvados/keep.py#L368

I think we want to add something like

```
curl.setopt(pycurl.CAINFO, ca_certs_path())
```

You'll want to do the same thing in put()

https://github.com/arvados/arvados/blob/master/sdk/python/arvados/keep.py#L457

pjotrp:
```python
# The following works by pulling in a new certificate file,
import certifi
curl.setopt(pycurl.CAINFO,certifi.where())
```

but I think ignoring system settings it is a poor way of doing it. I'll send a different fix.

pjotrp:
I also find there is a flag ARVADOS_API_HOST_INSECURE which is not recommended.

pjotrp:
@tetron this should work. Note that certificate handling in general is a nice mess because everyone plugs it in different places and we have multiple environment variables to deal with it (I count 5 at least). I avoid the python-certifi solution that is the common 'go it alone' approach because nothing ever works.

sdk/python/arvados/keep.py
```diff
@@ -375,6 +375,8 @@ def get(self, locator, method="GET", timeout=None):
                     curl.setopt(pycurl.HEADERFUNCTION, self._headerfunction)
                     if self.insecure:
                         curl.setopt(pycurl.SSL_VERIFYPEER, 0)
+                    else:
+                        curl.setopt(pycurl.CAPATH,os.path.dirname(arvados.util.ca_certs_path()))
```
tetron, line 379: Shouldn't this be CAINFO?

https://curl.haxx.se/libcurl/c/CURLOPT_CAPATH.html
"string naming a directory holding multiple CA certificates to verify the peer with."

https://curl.haxx.se/libcurl/c/CURLOPT_CAINFO.html
"string naming a file holding one or more certificates to verify the peer with."

pjotrp:
CAINFO points to a PEM. It did not work - I tried that first. CAPATH does the job simply pointing to the root of the certificate folder. Like I wrote, dealing with certificates is a mess.

tetron:
I just tried it with CAINFO and pointing to a single certificate bundle file (concatenated base64 encoded pem files), and it worked the way I expected it to.  I'm going to merge it using CAINFO because the behavior has to be consistent and searching the whole directory could lead to unexpected results.

PR #132, opened 2020-06-16
Remove debian tempfile and use mktemp instead

golharam:


cure:
Thank you for the contribution!

PR #135, opened 2020-07-28
Update arvados_config.rb

medcelerate:
Added fix to prevent ruby from failing when @ appears in the user for the database config.

### DIFF
--- a/services/api/config/arvados_config.rb
+++ b/services/api/config/arvados_config.rb
@@ -279,8 +279,8 @@ end
 # section of the config from database.yml
 #
 database_url = "postgresql://#{CGI.escape $arvados_config["PostgreSQL"]["Connection"]["user"]}:"+
-                      "#{$arvados_config["PostgreSQL"]["Connection"]["password"]}@"+
-                      "#{dbhost}/#{$arvados_config["PostgreSQL"]["Connection"]["dbname"]}?"+
+                      "#{CGI.escape $arvados_config["PostgreSQL"]["Connection"]["password"]}@"+
+                      "#{dbhost}/#{CGI.escape $arvados_config["PostgreSQL"]["Connection"]["dbname"]}?"+
                       "template=#{$arvados_config["PostgreSQL"]["Connection"]["template"]}&"+
                       "encoding=#{$arvados_config["PostgreSQL"]["Connection"]["client_encoding"]}&"+
                       "collation=#{$arvados_config["PostgreSQL"]["Connection"]["collation"]}&"+
--- a/services/api/config/arvados_config.rb
+++ b/services/api/config/arvados_config.rb
@@ -278,7 +278,7 @@ end
 # For config migration, we've previously populated the PostgreSQL
 # section of the config from database.yml
 #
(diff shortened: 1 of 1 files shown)


tetron:
Thanks @medcelerate !

Oh, I see what's going on here.  It is constructing a URL with all the database information and the username is not being URL quoted.

Two things:

1. We need [Developer Certificate of Origin](https://dev.arvados.org/projects/arvados/wiki/Developer_Certificate_Of_Origin) on the PR commit

1. A more general solution would be to use [CGI.escape](https://ruby-doc.org/stdlib-2.5.3/libdoc/cgi/rdoc/CGI/Util.html#method-i-escape)

medcelerate:
I thought about the second point over the weekend and yes that would be a better approach.
Where does 1 go? In the code or in the commit message?

tetron:
In the commit message

tetron:
Might be a good idea to squash it, we have a fussy script that checks that it is present on every commit not just the last one

services/api/config/arvados_config.rb
```diff
@@ -270,25 +271,22 @@ def arrayToHash cfg, k, v
...
-end
 
 #
 # If DATABASE_URL is set, then ActiveRecord won't error out if database.yml doesn't exist.
 #
 # For config migration, we've previously populated the PostgreSQL
 # section of the config from database.yml
 #
-ENV["DATABASE_URL"] = "postgresql://#{$arvados_config["PostgreSQL"]["Connection"]["user"]}:"+
+database_url = CGI.escape("postgresql://#{$arvados_config["PostgreSQL"]["Connection"]["user"]}:"+
```
tetron, line 281: I don't think that is going to work, that would escape the `@` and `:` characters that actually need to be there.  Should go inside the string interpolations:

```
database_url = "postgresql://#{CGI.escape $arvados_config["PostgreSQL"]["Connection"]["user"]}:"+
```

medcelerate: That would make sense, sorry this is my first foray into ruby.

tetron: No worries, we are very happy for the contribution!

services/api/config/arvados_config.rb
```diff
@@ -278,7 +278,7 @@ def arrayToHash cfg, k, v
 # For config migration, we've previously populated the PostgreSQL
 # section of the config from database.yml
 #
-database_url = CGI.escape("postgresql://#{$arvados_config["PostgreSQL"]["Connection"]["user"]}:"+
+database_url = "postgresql://#{CGI.escape $arvados_config["PostgreSQL"]["Connection"]["user"]}:"+
```
tetron, line 281: This ought to solve your issue but since we're at it would probably be worth adding CGI.escape to `password` and `dbname` just to be on the safe side.

medcelerate: Ok let me add that.

tetron:
Thanks!  I will merge it shortly

medcelerate:
Yay my first merge into arvados.

tetron:
Merged 4da1edec9c50c7adef8ab9fa500d5f3c40fb296f

PR #136, opened 2020-09-23
16867 - Rust SDK

andy-thomason:
This is a placeholder for development work on the Aravdos Rust API.

I have tried to include any links and resources we may need along the way. Please be
gentle as I am a first-timer in the Arvados community.


andy-thomason:
We now have pretty thorough coverage of the API.

To test the client api run `cargo test` in that directory. It should build
the api from arvdos-api.json into arvados-api.rs.

The tests are basic at present.  I would like to generate tests for all the endpoints
that would run on a real arvados server as well as the built-in mock server
in the client tests.

Using the API is pretty simple. For example:

```rust
    let arvados = ArvadosApi::new(arv_api_host.as_ref(), arv_api_token, arv_api_host_insecure)?;
    let resp : KeepService = arvados.keep_services().get("xyz".to_string()).fetch().await?;
    let resp : KeepService = arvados.keep_services().delete("xyz".to_string()).fetch().await?;
    let resp = arvados.keep_services().list().filters(Some(vec!["1".into(), "2".into()])).fetch()?;
```

I'd appreciate some help making real test cases.

We need to do some specials for data fetches and uploads and the "list" methods could
benefit from some internal logic to stream pages from the server.

Everything is async. Rust takes its cues from JavaScript which fully embraces the async
methodology. Attention needs to be paid to latency - ie. avoiding round trips to the server.

sdk/rust/arvados-client/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+type ArvadosError = Box<dyn std::error::Error + Send + Sync>;
+pub type Result<T> = std::result::Result<T, ArvadosError>;
+
+
+impl ArvadosClient {
+    pub fn new(arv_api_host: &str, arv_api_token: &str, arv_api_host_insecure: bool) -> Result<Self> {
+        let proto = if arv_api_host_insecure { "http" } else {"https" };
```
tetron, line 41: Not quite right.  The API server is always https.  The "insecure" flag means you want to accept self-signed certificates.

sdk/rust/arvados-client/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+        let base_url = format!("{}://{}/arvados/v1/", proto, arv_api_host);
+        let mut headers = HeaderMap::new();
+        let auth = format!("OAuth2 {}", arv_api_token);
```
tetron, line 44: Use "Bearer" instead of "OAuth2"

tetron:
Since you're using a code generation approach, you probably don't need to write a test for every separate API call.  The encoding/decoding of the API calls is entirely mechanical, so you just need to ensure there's a test case covering each of the various data types that can appear in the request / response fields.

You should be aware that the discovery document includes a bunch of deprecated APIs and some aliased methods.  Refer to these comments in the R SDK code generation about what you should filter out:

https://github.com/arvados/arvados/blob/2abdbcf641c2c6cb14764a3ed0d83849a4c7736a/sdk/R/R/autoGenAPI.R#L21-L23

https://github.com/arvados/arvados/blob/2abdbcf641c2c6cb14764a3ed0d83849a4c7736a/sdk/R/R/autoGenAPI.R#L128-L132

Arvados test orchestration happens in the bash script in `arvados/build/run-tests.sh`.  Here's a sketch of how to add it to the script:

```
testfuncargs["sdk/rust"]="sdk/rust"

install_sdk/rust() {
# install dependencies, generate the client from the API schema,
# any other commands that need to happen before tests can be run
}

test_sdk/rust() {
# actually run the test
}

install_all() {
  ...
  do_install sdk/rust
  ...
}

test_all() {
  ...
  do_test sdk/rust
  ...
}
```

One of the things that `run-tests.sh` does for you is start an API server in test mode, so you can actually write integration tests that run against a real API server with known test fixtures.

To be able to run all that you will need the rest of the development environment, see here https://dev.arvados.org/projects/arvados/wiki/Hacking_prerequisites

I recommend using a VM or Docker environment.  You have the added complication that the code we've provided to set up a development environment from scratch doesn't include the Rust compiler yet.

tetron:
To access Keep, you have two options.  You can use the WebDAV service(*), or interact with keepstore/keepproxy directly.  The WebDAV service is much easier to use but adds a layer of abstraction between the client and the keepstore.  A "high performance" client will interact with the keepstores directly, but needs to have all the logic to correctly divide up and reassemble files in a collection manifest, which is significantly more complicated.

(*) We're also adding support for an S3-compatible API in the next release, but it still operates at the same level of abstraction as WebDAV.

tetron:
By the way, in case you haven't seen this yet, we are starting up a user group video chat.  First meeting will be in a couple of weeks.  Details here:

https://forum.arvados.org/t/arvados-user-group-video-chat/47

sdk/rust/arvados-client/src/lib.rs
andy-thomason: Done.

sdk/rust/arvados-client/src/lib.rs
andy-thomason: Done. But I've kept http for testing as https is a bit over the top.

andy-thomason:
Thanks, for all the advice, @tetron . Next up some integration tests as you suggest.

andy-thomason:
I've added the chat to my video calendar. Thanks for having this in the day for Oxford.
I've been running a Discord chat room for some of my other projects.

andy-thomason:
I've moved the branch to a private repository to comply with Genomics open source policy.

PR #140, opened 2020-11-18
Fix provision.sh for salt stack deployment

pvanheus:
In the provision.sh script that uses salt stack to setup Arvados, there is a [reference](https://github.com/arvados/arvados/blob/8876199fb07f53263cd431a8a5cb15f4b9444361/tools/salt-install/provision.sh#L187) to `./${CONFIG_DIR}` but this happens at a time when the script has already [changed directory](https://github.com/arvados/arvados/blob/8876199fb07f53263cd431a8a5cb15f4b9444361/tools/salt-install/provision.sh#L170) to a different location. This patch saves the script directory before that change and uses that path instead of `.`.

### DIFF
--- a/tools/salt-install/provision.sh
+++ b/tools/salt-install/provision.sh
@@ -50,6 +50,9 @@ VERSION="latest"
 
 set -o pipefail
 
+# capture the directory that the script is running from
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
 usage() {
   echo >&2
   echo >&2 "Usage: $0 [-h] [-h]"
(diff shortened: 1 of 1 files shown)


javierbertoli (changes requested):
Hi @pvanheus, you're right, nice catch!

We need [Developer Certificate of Origin](https://dev.arvados.org/projects/arvados/wiki/Developer_Certificate_Of_Origin) on the PR commit, can you amend it so I can merge approve and merge it?

javierbertoli:
@pvanheus, I'll close this PR as I cherry-picked it [as part of a bigger](https://github.com/arvados/arvados/commit/cc9b9dac7be9c0bbf1d8e227466e9c71a7b3d35e) to fix the provision scripts. 

Thanks for your help!

PR #143, opened 2020-11-26
Fix run-test.sh to check for correct admin user

pvanheus:
Current code looks for a user with email admin@arva2.arv.local - this PR fixes that.


javierbertoli:
Thanks, @pvanheus !

PR #170, opened 2022-07-08
Use config parameter to fetch api token in KeepClient

MajewskiKrzysztof:
Arvados-DCO-1.1-Signed-off-by: Krzysztof Majewski majewski.kielce@gmail.com


MajewskiKrzysztof:
Using final parameter for API token is causing issues when trying to use the `KeepClient` as a Spring bean.

PR #182, opened 2022-12-20
Add warning notice that Composer is no longer maintained

Senui:
Since Arvados Composer is no longer supported / maintained it would help new users to the project to know that the installation instructions are no longer up to date (e.g. the `arvados-composer` package is no longer part of the package managers' repositories). Keeping the installation page up might be useful for anyone who might want to pick this up in the future.


brettcs:
The doc has simply been removed since (in da5858d5f794c14cf00b830166bb34b1bcd79ba5).

PR #247, opened 2024-06-19
Support setting replication level at destination in arv-copy options

zoe-translates:
arv-copy effectively used a hard-coded replication level 2 for the copied collections at the destination, bypassing the default replication level set in the destination cluster's configuration file. There was no command-line option to override this behavior.

A new command-line option, --replication, is added to the arv-copy command, following the arv-put command's semantics. If left unspecified, the destination's default replication setting is used. If that setting cannot be found, use the fallback value of 2.

Arvados-DCO-1.1-Signed-off-by: Zoë Ma <zoe.ma@curii.com>


brettcs (changes requested):

brettcs:
This was merged as part of 53b3450172. Thank you!

PR #249, opened 2024-07-03
Python and CWL command line tools: More robust handling of storage-class options

zoe-translates:
For CLI clients (`arv-copy`, `arv-put`, `arvados-cwl-runner`), the user can specify the storage classes of output collections with options like `--storage-classes`, by passing the names of the classes as a comma-separated string.

We make the parsing of such arguments more resilient against invalid (e.g. all-blank) and duplicated input values. Specifically, the input string is split at the commas, and each segment is stripped of leading and trailing space characters. Then, empty values are discarded and duplicate values removed, with only the first appearances kept. (This is also reused for processing `--varying-url-params` etc.)

Moreover, for the CWL client in particular, we no longer use the hard-coded string `"default"` as the default value to the `--storage-classes` and `--intermediate-storage-classes` options. Instead, if the user does not explicitly specify any (valid) values (and none are specified in the CWL file's hints or requirements section), we gather and use the info about the real "default storage classes" as they are actually configured on the cluster where each container is to be run.

Internally, the `arvados_cwl.context.ArvRuntimeContext` object now stores the `storage_classes` and `intermediate_storage_classes` attributes as lists rather than strings. This, together with the decoupling of "using default classes" with "being equal to the magic string `"default"`", tends to simplify their processing in code elsewhere.

For the common use cases where the storage class "default" is indeed the only default storage class, the changes introduced here won't have any observable change of behavior.

### DIFF
--- a/sdk/cwl/arvados_cwl/__init__.py
+++ b/sdk/cwl/arvados_cwl/__init__.py
@@ -29,6 +29,7 @@ import arvados.config
 import arvados.logging
 from arvados.keep import KeepClient
 from arvados.errors import ApiError
+from arvados.util import csv_to_list
 import arvados.commands._util as arv_cmd
 
 from .perf import Perf
@@ -183,9 +184,9 @@ def arg_parser():  # type: () -> argparse.ArgumentParser
     parser.add_argument("--enable-dev", action="store_true",
                         help="Enable loading and running development versions "
                              "of the CWL standards.", default=False)
-    parser.add_argument('--storage-classes', default="default",
+    parser.add_argument('--storage-classes', default="",
                         help="Specify comma separated list of storage classes to be used when saving final workflow output to Keep.")
-    parser.add_argument('--intermediate-storage-classes', default="default",
+    parser.add_argument('--intermediate-storage-classes', default="",
                         help="Specify comma separated list of storage classes to be used when saving intermediate workflow output to Keep.")
 
     parser.add_argument("--intermediate-output-ttl", type=int, metavar="N",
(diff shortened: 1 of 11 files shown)


brettcs (changes requested):
* I don't think we should make `csv_to_list` part of our public API, so I think we should move it to our internal module `arvados.commands._util`. And in order to keep the private API a little more useful for future tools, I think it would be more useful if the signature was:
  ```python
   def str_to_list(arg: str, sep: str=',') -> List[str]: ...
   ```
   This way it could be used for parsing all kinds of list arguments no matter what their delimiter is.
* Let's use this function directly as the `type` for the relevant argparse arguments, so future developers don't have to remember to call it later.
* Similarly, since it needs to be part of our public API, let's make the signature of the storage classes function:
   ```python
   def iter_storage_classes(config: Dict[str, Any], filter_fn: Callable[[Dict[str, Any]], bool]=operator.methodcaller('get', 'Default'), fallback: str='default') -> Iterator[str]: ...
   ```
   This would provide a lot more flexibility to users without making things any more difficult for the simple case. When you need a list you can just wrap it in the `list` function.
* Please add unit tests specifically for these two new functions. This way, we can test a wider variety of inputs with minimal overhead. If you like, you can write these tests in a more modern "pytest" style instead of the old unittest style. `tests/test_internal.py` has a basic example.
* For the CWL tests that check that the parsing falls back to default storage classes, please add an assertion to check that this is done. If this is impractical, I think it would be better to simply remove the tests. If we have unit tests for the `type` function plus one test to confirm that it is used, the additional overhead of invoking multiple CWL runners without any additional assertions beyond "it ran" does not seem worth it to me.
* Following PEP 257, please make sure the first line of the docstring is a single physical line. The next line in the source should be a blank line.
* Following our Markdown style, you do not need to further indent the bullets for arguments. It would be better if you didn't to avoid the risk that it might get interpreted as a Markdown code block.
* The oldest version of Python we support is 3.8. Because of that, we cannot subscript built-in types like `list[str]` or `dict[str, Any]` yet. Please update these annotations to use the titlecase versions from the `typing` module.

brettcs:
This was merged as of 53b3450172. Thank you!

PR #250, opened 2024-07-08
21989: Keepstore: Use comma separator for X-Keep-Storage-Classes-Confirmed header

zoe-translates:
When generating the X-Keep-Storage-Classes-Confirmed header, use comma (",") instead of semicolon (";") as the separator for multi-entry value.

This is to be consistent with keep-proxy's behavior and the arv-put client's expectation. Plus, it follows RFC 9110 (see sec. 5.2).

Without the fix, arv-put wrongly believes that storage classes are not supported on the cluster it is putting to. The spurious warning "X-Keep-Storage-Classes header not supported by the cluster" can be seen in arv-put messages and in Workbench2.


tomclegg:
Thank you for the report & fix!

PR #252, opened 2024-07-11
21998: properly percent-encode paths in keep-web directory listing page

zoe-translates:
Use percent-encoded form of the relative URLs as the value of "href" attribute in directory listing page generated by keep-web

The "wget" command example shown on the page now has single-quotes around the URL argument.

For testing involving complex URL or path patterns, use the HTML parser provided by golang.org/x/net/html to scrape the directory listing page, instead of using regular expressions that may get unwieldy.

Minor edits to the HTML template for compliance and ease of testing.


tomclegg:
Thank you!

PR #265, opened 2024-09-27
Add UsePathStyle as a config parameter in S3VolumeDriverParameters

dunglam2k:
Added UsePathStyle in S3VolumeDriverParameters and set the default value to false in config.default.yml


tomclegg:
@dunglam2k thank you!

PR #296, opened 2025-04-08
Update Keep Web API Client URL Builder

MajewskiKrzysztof:
Arvados-DCO-1.1-Signed-off-by: Krzysztof Majewski <majewski.kielce@gmail.com>


MajewskiKrzysztof:
Hi @tetron 
There is a security concern on "c=uuid". This endpoint also returns data when opened in browser and is not domain dependent. 
Do you think it would be a good idea to call the `/config` endpoint and use the URL from `Services.WebDAVDownload` value  as a keep-web host?

Issue #371, opened 2026-04-15
Add arv create to arvcli.py


brettcs:
Add the `arv create RESOURCE` subcommand to `arvcli.py`. The basic flow is:

1. Create a temporary file in the specified format and with the given `owner_uuid` (if one is provided).  
2. Open the user’s editor on that temporary file and wait for it to finish.  
3. Parse the result and use it as the body for a `RESOURCE/create` API request.  
   1. If the parsing result is an empty object, the user wants to exit; do not make an API request.  
   2. If parsing or the API server return an error, add information about the error to the file and loop back to step 2\.  
   3. Otherwise, output the result of the API request in the user’s chosen format.

**Backwards compatibility:** Because this tool is inherently interactive, we do not need to be concerned that scripts and deployment tools depend on it. The only compatibility requirement is that we use the same command syntax, with all the same options having all the same meaning. It is okay to change how we invoke the editor, how we format messages, how we report errors, etc.

**Options:**

* `arv create` uses the global `--format` option to decide what format to use for the temporary file, and how to parse it. `json` and `yaml` are both supported. Calling `arv create --format=uuid` is an error.  
* For itself, `arv create` supports the standard `--help` option and one more: `--project-uuid`/`-p`. If this is set, then the initial temporary file that is created has this value as the `owner_uuid`. (The user still has the option to edit it from there, and if they do we respect their edits.)

**Invoking the editor:**  
Build a command list like this:

```python
if cmd_str := (os.environ.get('VISUAL') or os.environ.get('EDITOR')):
    cmd = shlex.parse(cmd_str)
elif cmd_str := shutil.which('nano'):
    cmd = [cmd_str]
else:
    cmd = ['vi']
```

Then append the tempfile filename to `cmd` and run it with `subprocess.run`.

Do not second-guess user configuration. If the user has `$VISUAL` or `$EDITOR` set to something that can't be run, report the error and abort. Do not fall back to the later steps.

**Future Development:** Later we will build `arv edit`. This has a similar flow, but it builds the initial tempfile by getting an object by UUID; then updates it afterwards with `RESOURCE/update`. As you consider breaking out your work into functions, a useful way to think about breaking things up is to think about which functions can be reused by `arv edit` (building the tempfile from an initial object; running the editor; parsing the result) and which cannot (building the initial object; the specific API request afterwards).

ZoeCMa:
[...]

PR #385, opened 2026-04-28
371: `arvcli.py create` subcommand

ZoeCMa:
I've been hacking away at #371 but it's not there yet. Specifically, this is not yet done:

> [ii.] If parsing or the API server return an error, add information about the error to the file and loop back to step 2.

In [this comment](https://github.com/arvados/arvados/issues/371#issuecomment-4335983480) I started a higher-level discussion about the user experience of this and related UX thoughts.

I see that the code I add so far is already a big changeset and I feel that some review could be helpful here, if simply to figure out whether course-correction is needed. Many thanks!

The code does the following:

* Added new classes, `JSONEditingProcess` and `YAMLEditingProcess`, that inherit from a common base that present a somewhat unified way of performing the object-editing with the user in the loop. The subclasses adapt to the idiosyncrasies of formatting, validation, etc. for the two file formats.
* Added `_handle_external_editor_command()` to implement the editing loop.
* Added tests and especially an executable script, `tests/editor_simulator.py`, that simulates the very basic aspects of an external editor in tests (with several modes of injecting faults, currently not used in our tests).


sdk/python/arvados/commands/arvcli.py
```diff
@@ -16,19 +16,42 @@
...
 class _ArgTypes:
     """Private namespace class for JSON-related CLI argument types."""
+    group_uuid_pattern = re.compile(r"\A[0-9a-z]{5}-j7d0g-[0-9a-z]{15}\Z")
+
+    @staticmethod
+    def group_uuid(text: str) -> str:
+        """Validate an Arvados group UUID as the value of a CLI argument (an
+        Arvados project being a type of group).
+        """
+        if _ArgTypes.group_uuid_pattern.match(text):
```
brettcs: Please reuse `arvados.util.group_uuid_pattern` here.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -313,6 +607,29 @@ def __init__(self, resource_dictionary, **kwargs):
...
+        for cli_name, resource in self._subcommand_to_resource.items():
+            resource_schema = self.resource_dictionary[resource]
+            if "create" in resource_schema.get("methods", {}):
+                creatable_targets.add(cli_name)
+        create_parser = subparsers.add_parser(
+            "create", help="Create Arvados object using external editor"
+        )
+        create_parser.add_argument(
+            "target_resource",
+            choices=sorted(list(creatable_targets)),
```
brettcs: `sorted` already returns a list, so calling `list` is redundant and unnecessary.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +734,145 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+        for k, v in vars(args).items()
+        if k not in ArvCLIArgumentParser.global_args
+    }
+
+    sys.exit(_call_resource_method(arv_method, method_args, args.format))
+
+
+def _handle_external_editor_command(api_client, parser, args) -> NoReturn:
+    if args.format == "uuid":
+        print(
```
brettcs: Prefer using `parser.error` here.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +734,145 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+    except OSError:
+        stdin_fileno = 0
+    # Refuse to run when we're not in an interactive session. Some editors may
+    # be unwilling to quit even when not attached to a terminal (e.g., vim),
+    # which would have caused us to wait without making progress. Others (e.g.,
+    # nano) may quit immediately with non-zero code, but editor exit codes are
+    # flaky and not well-documented or standardized (see
+    # https://stackoverflow.com/a/46678151,
+    # https://unix.stackexchange.com/a/293461), and we can't rely on them.
+    if not os.isatty(stdin_fileno):
```
brettcs: Prefer `sys.stdin.isatty()`. This should return a meaningful result even for test objects.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +734,145 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+        editing = JSONEditingProcess(initial_object=obj_stub)
+    else:
+        editing = YAMLEditingProcess(initial_object=obj_stub)
+
+    with editing:
+        while True:
+            try:
+                editing.edit()
+            except OSError as err:
+                print(
```
brettcs, line 755: Please streamline this message as 

```
f"error: failed to execute editor `{cmd_str}`: {err}"
```

where `cmd_str` is `' '.join(shlex.quote(s) for s in cmd_seq)`. This is more human-friendly than a list repr.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +734,145 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+        )
+        sys.exit(1)
+
+    obj_stub = {"owner_uuid": args.project_uuid} if args.project_uuid else None
+    if args.format == "json":
+        editing = JSONEditingProcess(initial_object=obj_stub)
+    else:
+        editing = YAMLEditingProcess(initial_object=obj_stub)
+
+    with editing:
```
brettcs, line 746: In this loop, I suggest we format all messages like this:

```
[prefix]: [message]: [details]
```

This follows conventions of our other tools better, avoids excess verbiage, and more importantly, it provides internal consistency. Some of these messages have the details first, and others at the end. I think formatting all messages the same way promotes internal consistency.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +734,145 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+                sys.exit(1)
+
+            try:
+                edited_obj = editing.load()
+            except EditingContentError as err:
+                # Invalid input from editor; emit error message and let the
+                # user try again.
+                print(str(err), file=sys.stderr)
+                print(
+                    "Error: likely syntactic error in the file being edited;"
```
brettcs: Suggest

```
f"error: invalid input: {err}"
```

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +734,145 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+                    " see error messages above.",
+                    file=sys.stderr
+                )
+                while (wants_retry := _ask_reedit()) is None:
+                    pass
+                if wants_retry:
+                    # NOTE: Back to the start of the "while True" loop!
+                    continue
+                sys.exit(1)
+            if edited_obj is None:
```
brettcs: Suggest just `if not edited_obj`. I see no reason not to treat an empty object the same as an empty file.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +734,145 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+    # https://unix.stackexchange.com/a/293461), and we can't rely on them.
+    if not os.isatty(stdin_fileno):
+        print(
+            "'create'/'edit' subcommands can only run interactively when"
+            " input/output are a terminal.",
+            file=sys.stderr
+        )
+        sys.exit(1)
+
+    obj_stub = {"owner_uuid": args.project_uuid} if args.project_uuid else None
```
brettcs: Suggest we make `{}` the default `obj_stub`, then exit if the user saves an empty object (see below).

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +734,145 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+                )
+                while (wants_retry := _ask_reedit()) is None:
+                    pass
+                if wants_retry:
+                    # NOTE: Back to the start of the "while True" loop!
+                    continue
+                sys.exit(1)
+            if edited_obj is None:
+                print(
+                    "Info: file content is empty or blank; stopping the"
```
brettcs:
```
"notice: input is empty; exiting without changes"
```

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +734,145 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+                " message above.",
+                file=sys.stderr
+            )
+            while (wants_retry := _ask_reedit()) is None:
+                pass
+            if wants_retry:
+                continue  # NOTE: Back to the start of the "while True" loop!
+            sys.exit(1)
+        # End of the "while True" loop for editor interaction.
+    sys.exit(0)  # End of the NoReturn function.
```
brettcs: I think exiting 0 has the potential to mask bugs. Getting here is not the expected flow of the function. If we got here, something went wrong, and we should indicate that.

I think a nicer way to approach this would be to tie the loop to the exit code. Then the loop condition can help ensure we have the intended flow:

```python
call_status = None
while call_status is None:
    ...
    call_status = _call_resource_method(...)
    if call_status != 0:
        print(error ...)
        wants_retry = None
        while wants_retry is None:
            wants_retry = _ask_reedit()
            if wants_retry:
                call_status = None
sys.exit(call_status)
```

sdk/python/arvados/commands/arvcli.py
```diff
@@ -231,6 +254,277 @@ def get_method_options(method_schema):
...
+    @abc.abstractmethod
+    def deserialize(self, file: TextIO) -> Any:
+        """Abstract method for loading from the file-like object `file` as
+        text. Returns the object deserialized from the text content.
+        """
+
+    def dump(self, obj: Any) -> None:
+        """Append the serialized object to the temporary file."""
+        if self.tmp_file is None or self.tmp_file.closed:
+            raise RuntimeError("Temporary file is not available for writing")
```
brettcs: This check-and-raise gets repeated several times. Please DRY out a method to do it. It can take the exception message as an argument.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -231,6 +254,277 @@ def get_method_options(method_schema):
...
+        # The following should not be done while the child process is pending.
+        pos = self.tmp_file.tell()
```
brettcs: Now that we've decided to do interactive prompting instead of putting errors in the tempfile, I believe you can just overwrite the file with the new contents and avoid this whole tell-and-seek dance. Is there something I'm missing?

sdk/python/arvados/commands/arvcli.py
```diff
@@ -231,6 +254,277 @@ def get_method_options(method_schema):
...
+        if self.tmp_file is None or self.tmp_file.closed:
+            raise RuntimeError("Temporary file is not available for reading")
+        # Snoop the file to see if it consists of only whitespace characters
+        # (including empty lines); if so, return the special value None to
+        # indicate that the user has abandoned the edit.
+        with open(self.tmp_file.name, "r") as fdup:
+            if not fdup.read().strip():
+                return None
+
+        pos = self.tmp_file.tell()
```
brettcs: If I'm right above we don't need tell-and-seek, then we don't need it here either. Every read method is going to start with `self.tmp_file.seek(0)` so there's no reason to remember and restore the current position.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -231,6 +254,277 @@ def get_method_options(method_schema):
...
+        if not isinstance(obj, Mapping):
+            raise self.input_error_type(
+                path=path,
+                # NOTE: the "official" term for the curly-braced entity is
+                # "object" (see
+                # https://datatracker.ietf.org/doc/html/rfc7159.html,
+                # https://www.json.org/json-en.html)
+                # but that word may be a little confusing, so we elaborate a
+                # bit in the user-facing message.
+                original_exception=ValueError(
```
brettcs, line 460: IMO, for both formats, we should write this error like:

```
f"[format] input has type {type(obj).__name__}, not a valid Arvados object"
```

This has two benefits. First, while I respect you looked up and used the appropriate terminology for each format, we can't expect our users to have done the same. Therefore it seems ideal if we can avoid format-specific terminology altogether. Second, this message tells the user what type of object we parsed, which might be a clue as to what was wrong with the input.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +734,145 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+    # Refuse to run when we're not in an interactive session. Some editors may
+    # be unwilling to quit even when not attached to a terminal (e.g., vim),
+    # which would have caused us to wait without making progress. Others (e.g.,
+    # nano) may quit immediately with non-zero code, but editor exit codes are
+    # flaky and not well-documented or standardized (see
+    # https://stackoverflow.com/a/46678151,
+    # https://unix.stackexchange.com/a/293461), and we can't rely on them.
+    if not os.isatty(stdin_fileno):
+        print(
+            "'create'/'edit' subcommands can only run interactively when"
```
brettcs: Please add an "error:" prefix.

brettcs (changes requested):
These comments are just on the implementation, I still need to go over the tests.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: In 132cd2fb5e41586cab7c7a108c54d47a6eb2c36a it's done here and elsewhere.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: Fixed in 102b2f2c4c2368ba11c182f006cd5201e48de8f3.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: Fixed in 102b2f2c4c2368ba11c182f006cd5201e48de8f3.

sdk/python/tests/editor_simulator.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
brettcs: All these cases are worth covering and I appreciate that you did so. However, I believe we can implement them all with standard Unix tools, and don't need to maintain our own utility like this to do so. I think we just need one utility script `writefile.sh` to help us overwrite/append files:

```sh
#!/bin/sh
srcfile="$1"; shift
exec tee "$@" <"$srcfile" >/dev/null
```

With that, here are the tools we can use for different cases:

| Your flag | `$EDITOR` value |
|---|---|
| `--input SRC` | `writefile.sh SRC` |
| `--append` | `writefile.sh SRC -a` |
| `--replace` | `cp SRC` |
| `--delete` | `rm` |
| `--crash` | `false` |

To implement `--next`, we can have the test overwrite `SRC` when the code under test prompts for a retry. Since we have to mock stdin anyway, that's a clear place we can hook in: we can rewrite the file any time our mock stdin is read.

sdk/python/tests/test_arvcli.py
```diff
@@ -3,21 +3,59 @@
...
+    buf = open(tmp_path / "mock_stdin", "w+")
+    monkeypatch.setattr(sys, "stdin", buf)
+    try:
+        yield sys.stdin
+    finally:
+        buf.close()
+
+
+@pytest.fixture
+def aux_client():
```
brettcs: I don't understand what "aux" is referring to in this name. I suggest we just name this `arv_client`. It's generic enough that we can define it in `conftest.py` for other tests to use.

sdk/python/tests/test_arvcli.py
```diff
@@ -643,3 +692,339 @@ def test_current(self, run_arvcli):
 
     # TODO: investigate possible authorization issue with testing
     # the create_system_auth method.
+
+
+GEC = arvcli.ObjectEditingProcessBase.get_editor_cmdline
```
brettcs: It would follow conventional style better to define this as a class variable `gec` under `TestGetEditorCmdline`.

sdk/python/tests/test_arvcli.py
```diff
@@ -643,3 +692,339 @@ def test_current(self, run_arvcli):
...
+    assert not sr.called
+
+
+def yaml_dumps(obj) -> str:
+    buf = io.StringIO()
+    yaml.dump(obj, stream=buf)
+    return buf.getvalue()
+
+
+EditFormatCase = namedtuple("EditFormatCase", ("format", "dumps", "loads"))
```
brettcs: I think this would be more straightforward if the garbage text was an attribute of `EditFormatCase`. That text is necessarily format-specific, so it makes sense to tie them together. It's okay if not every test around `EditFormatCase` uses the garbage text.

sdk/python/tests/test_arvcli.py
```diff
@@ -643,3 +692,339 @@ def test_current(self, run_arvcli):
...
+FORMAT_CASES = (
+    EditFormatCase("json", json.dumps, json.loads),
+    EditFormatCase("yaml", yaml_dumps, yaml.load)
+)
+GARBAGE_TEXTS = (
+    "foo: bar foo: bar",  # invalid JSON
+    "{{}}"  # invalid YAML
+)
+EDITOR_INPUT_OBJ = {"name": "a new project", "group_class": "project"}
```
brettcs: It would be nicer to make this a fixture that returns the object in order to make sure that no test accidentally mutates it and interferes with other tests.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: Fixed in 49f981718204a8c385167404b5a88d2c5e7f103f.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: I simplified this loop in cf980a7ead00d036efa38d793c3922ab72f57399 where the error messages that doesn't add new info are removed. `EditingContentError.__str__()` and `_call_resource_method()` already largely follow the same convention here.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: This is done in 00519971fe79bfa4bc7d5db4292af5b1917b82e5.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: Done in b43399b7b47c7accdbea9191fb31d0ab5a42749c.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: Done in b43399b7b47c7accdbea9191fb31d0ab5a42749c.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: Done in b43399b7b47c7accdbea9191fb31d0ab5a42749c.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: Done in b43399b7b47c7accdbea9191fb31d0ab5a42749c.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: Done in b43399b7b47c7accdbea9191fb31d0ab5a42749c.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: This is done in ae83779ba0742bb7262f090e44bc4e935d5c3b0e.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: This is done in 67671bf7f40b5c83848aa0e273448796a2a82a23.

sdk/python/tests/test_arvcli.py
ZoeCMa: This is done in 3493175a7aaa2fb955e935c0466db5eec78b2f20.

Not yet in `conftest.py` yet but I agree there's a lot we can put there.

BTW the `aux` was meant to indicate 'auxiliary', as in 'additional client other than the one underneath `arvcli.py`', but it turned out confusing.

sdk/python/tests/test_arvcli.py
ZoeCMa: This is done in 3493175a7aaa2fb955e935c0466db5eec78b2f20 too.

sdk/python/tests/test_arvcli.py
ZoeCMa: Done in ad05bfc13601168cf20f69edeedba22d72f29791. This is a great idea.

sdk/python/tests/editor_simulator.py
ZoeCMa: Thank you for this idea. It is implemented in 2bcf097671819298407c85483387d492e463f04a, a fairly big diff. Overall, we have now more flexibility because, in a mocked sense, _each_ `"y"` choice by the user at the prompt can be immediately followed by an arbitrary "edit".

sdk/python/tests/test_arvcli.py
ZoeCMa: This is done in 59881358afecd612c11211152b0b40cf024de2b2, 446f298781570cd5bb6fda3525eb6e2492637108.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -231,6 +253,256 @@ def get_method_options(method_schema):
...
+        """Perform a basic sanity check for the temp file being usable."""
+        if self.tmp_file is None or self.tmp_file.closed:
+            raise RuntimeError("Temporary file is not available")
+
+    def dump(self, obj: Any) -> None:
+        """Overwrite the temporary file with the serialized object `obj`."""
+        self.check_tmp_file()
+        # The following should not be done while the child process is pending.
+        self.tmp_file.seek(0)
+        self.tmp_file.truncate()
```
brettcs: You can just say `self.tmp_file.truncate(0)`.

sdk/python/arvados/commands/arvcli.py
brettcs: Today I learned about `shlex.join`, that's way nicer, thanks!

sdk/python/arvados/commands/arvcli.py
```diff
@@ -425,9 +712,109 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
...
+
+    sys.exit(_call_resource_method(arv_method, method_args, args.format))
+
+
+def _handle_external_editor_command(api_client, parser, args) -> NoReturn:
+    obj_stub = {"owner_uuid": args.project_uuid} if args.project_uuid else {}
+    if args.format == "json":
+        editing = JSONEditingProcess(initial_object=obj_stub)
+    else:
+        editing = YAMLEditingProcess(initial_object=obj_stub)
```
brettcs: Please verify that `args.format == 'yaml'` and raise a RuntimeError or similar if `args.format` is an unexpected value. This helps ensure we address all cases if we add another format in the future.

sdk/python/tests/test_arvcli.py
```diff
@@ -643,3 +692,390 @@ def test_current(self, run_arvcli):
...
+        with open(edit_source, "w") as s:
+            # Note that not all actions uses the edit_source file.
+            s.write(content)
+        sep = "-----\n"
+        match action:
+            case "replace":
+                logf.write(f"replace tmpfile: {sep}")
+                logf.write(content)
+                logf.write(sep)
+                editor_cmd = "cp {edit_source}"
```
brettcs: Here and in other cases, you should `shlex.quote(edit_source)` to help ensure it's parsed correctly.

sdk/python/tests/test_arvcli.py
```diff
@@ -643,3 +692,390 @@ def test_current(self, run_arvcli):
...
+                editor_cmd = "rm"
+            case "crash":
+                logf.write(f"crash: {sep}")
+                editor_cmd = "false"
+            case "append":
+                logf.write(f"append to tmpfile: {sep}")
+                logf.write(content)
+                logf.write(sep)
+                editor_cmd = f"{writefile_script} {edit_source} -a"
+            case _:  # The most common case: fill target file with content.
```
brettcs: To prevent surprising behavior if the caller passes a typo value, please check for an explicit string here, then raise a RuntimeError or similar in the case of an unrecognized string.

Whatever string you choose here, you can make the default value for the `action` argument. That keeps the convenience for callers doing this common case.

sdk/python/tests/test_arvcli.py
```diff
@@ -643,3 +692,390 @@ def test_current(self, run_arvcli):
...
+    base_dir = tmp_path / "editor_input"
+    base_dir.mkdir()
+    # Temporary file that gets written each time the editor is "installed"
+    edit_source = base_dir / "source"
+    # Persistent file that keeps a record of editor input, for each request to
+    # this fixture (typically means function scope).
+    log = base_dir / "log"
+    logf = open(log, "a")
+
+    def editor_fcn(content: str = "", action: str = ""):
```
brettcs: Small style thing, prefer the name `editor_func`. We use the `_func` suffix elsewhere in the codebase.

sdk/python/tests/test_arvcli.py
```diff
@@ -643,3 +692,390 @@ def test_current(self, run_arvcli):
...
+    EditFormatCase("yaml", yaml_dumps, yaml.load, "{{}}")
+)
+
+
+@pytest.fixture
+def new_project():
+    return {"name": "a new project", "group_class": "project"}
+
+
+class AnswerAndEdit:
```
brettcs: This class reimplements functionality available from `unittest.mock.Mock`. It would be better to use that rather than implement our own. You can use the [`side_effect` attribute](https://docs.python.org/3.10/library/unittest.mock.html#unittest.mock.Mock.side_effect) to return different values across calls or write a new source file. You can use the class' different assert methods to check it was called the way you expect.

brettcs (changes requested):

sdk/python/arvados/commands/arvcli.py
ZoeCMa: This is done in 683893026011ecb7d5a773f818370a3a9afe6a69.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: This is done in 55cf6caec2a33f9ae0350684c64912e3c91e03d0.

Also in 1dad03d8926f9c6a1f269fd008f85d900be0abeb, allow the user to say `--format=YAML` instead of having to use all-lowercase `yaml`.

sdk/python/tests/test_arvcli.py
ZoeCMa: I did that, in c44c4646df24e70e1c0d3e7cae86032d70de5df3. Arguably in a more verbose way (writing all the underlying commands as lists and then calling `shlex.join()` once) ... kinda liked the idea of `join()` as the inverse of `split()` we used in the main code.

sdk/python/tests/test_arvcli.py
ZoeCMa: Thanks, did that in 0a5444c297f2d740bbf6ab27745d9dd1878eaefa.

sdk/python/tests/test_arvcli.py
ZoeCMa: Thanks, good catch, done in 944a6192212e8ba20932e73cc6ad7aa826ec688d.

sdk/python/tests/test_arvcli.py
ZoeCMa: I'm actually kinda tempted to switch the places of two arguments to the function, so that `action` goes before the `content`, and so I can say `setup_editor("write", something)` in the code.

This would also make it less clunky to set an editor action that doesn't take content, such as `setup_editor("delete")` as opposed to `setup_editor("", "delete")` or so. But we don't have those tests yet -- all we have done is `"write"` and `"append"`.

The drawback is that the most common case (`"write"`) would have to be spelled out.

Thoughts? Thanks.

brettcs: I agree that `(action, content)` seems like the more intuitive argument order.

You could still have the signature be `(action='write', content='')`, and then tests in the common case could say `setup_editor(content='foo')`.

That doesn't really save typing, but IMO an API that's intuitive and easy to remember saves more time than a few keystrokes.

sdk/python/tests/test_arvcli.py
ZoeCMa: The wheel is un-reinvented in 45becc0e554878e4c86546b68bbb38a4296aa7ce. In subsequent 51b9b073293fa67868b326228e425a6edc9fbbcb and 2d8a0f47f965ff2926c9f00fa3f1b333d81a751d I also re-named the context manager that limits the scope of the builtin-function-monkeypatching (two commits due to typo).

sdk/python/tests/test_arvcli.py
ZoeCMa: Thanks, this is done in 39d96e2f28d0e5f85eca7e81fbc382f28710b0bb.

ZoeCMa:
https://ci.arvados.org/job/developer-run-tests/5081/

  * This is a new feature PR (see #371, including the comments). Compared with the older design in `arv create` (the Ruby client), the most notable is that we don't put lines starting with `#` in the tempfile as a means of providing feedback to the user. We keep the temp file "clean" and under the control of the user's text editor. The feedback is printed to the terminal which pauses at errors, waiting for user to ack them and proceed interactively.
  * N/A
  * Manual test involves running `arvcli.py create` against a test cluster, with JSON and YAML format options, while setting `$VISUAL` and `$EDITOR` to various values.
 * Actually, the branch is not based on the 'most recent' main, but the diverging main has no other changes pertaining to the `sdk/python` subdir.
  * N/A
  * Currently, there are `--help` messages and inline docstrings; full documentation (web pages, and maybe man pages) for the new `arvcli.py` would have to be "ported" too.
  * N/A
  * The types of resource to be `create`d are read from the discovery document served by the API, and we don't hard-code resource-specific logic in the client.
  * Yes

PR #389, opened 2026-05-08
WIP: arvcli.py help message improvements.

ZoeCMa:
* For external, "pass-through" commands such as `keep docker`, use `add_parser(add_help=False)` for their subparser instances alone, rather than forcing the "sans-help" parser class on all subparsers via `add_subparsers(parser_class=partial(..., add_help=False))`. This removes the problem for the non-passthrough subcommands by restoring the normal behavior of subparsers.
* Remove the custom workaround for this problem we created ourselves.
* When adding a subparser, populate the `help=` keyword argument, so the help message of its parent command properly displays the subcommands.
* When a subcommand is called without required "sub-subcommands", for example `arvcli.py collection` without any method sub-subcommand, dump the full help showing actual available methods and their summaries, rather than just showing the default message saying `error: the following arguments are required: method`. This is done by a custom subclass of `argparse.ArgumentParser` that modifies `error()` method behavior.
* More consistent error messaging for unrecognized parameter in resource-method call.

TODO: the `%(prog)` part in the usage message shown for external subcommands, such as `arvcli.py keep docker --help` or `arvcli.py keep docker --invalid-parameter`, is incorrect; it says `arvcli.py` but should include the full subcommand.

### DIFF
--- a/sdk/python/arvados/commands/arvcli.py
+++ b/sdk/python/arvados/commands/arvcli.py
@@ -863,12 +863,12 @@ def dispatch(arguments=None):
     # required by the parser. But "method" may be absent, as is in the case of
     # external commands like "ws" or "copy".
     method = getattr(args, "method", "")
+    command_key = f"{args.subcommand} {method}" if method else args.subcommand
 
     # Are we calling an external command?
-    ext_module = cmd_parser.external_command_modules.get(
-        f"{args.subcommand} {method}" if method else args.subcommand
-    )
+    ext_module = cmd_parser.external_command_modules.get(command_key)
     if ext_module is not None:
+        sys.argv[0] = f"arv {command_key}"
         _handle_external_command(ext_module, remaining_args)  # Exits.
 
     # Are we doing an API resource call?
@@ -881,7 +881,7 @@ def dispatch(arguments=None):
             cmd_parser.error(
                 f"unrecognized arguments: {', '.join(remaining_args)}\n"
                 f"Try: {sys.argv[0]} --help\n"
-                f"     {sys.argv[0]} {args.subcommand} {method} --help",
+                f"     {sys.argv[0]} {command_key} --help",
                 with_help=False
             )  # Exits with status 2.
         _handle_resource_method(api_client, resource, args)  # Exits.
--- a/sdk/python/arvados/commands/arvcli.py
+++ b/sdk/python/arvados/commands/arvcli.py
@@ -501,7 +501,22 @@ class YAMLEditingProcess(ObjectEditingProcessBase):
         return obj
 
 
-class ArvCLIArgumentParser(argparse.ArgumentParser):
+class FullHelpOnErrorArgumentParser(argparse.ArgumentParser):
+    """Argument parser subclass that customizes the `error()` method.
+
+    Intended to be used as a base to a parser with complex subparsers, to print
+    more-useful information when a required subcommand is missing.
+    """
+    def error(self, message, with_help=True):
+        if with_help:
+            self.print_help(sys.stderr)
+            print(file=sys.stderr)
+        # XXX: Use of self.prog, undocumented.
+        print(f"{self.prog}: error: {message}", file=sys.stderr)
+        sys.exit(2)
+
+
+class ArvCLIArgumentParser(FullHelpOnErrorArgumentParser):
     """Argument parser for `arv` commands.
     """
     global_args = frozenset((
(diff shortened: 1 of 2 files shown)


brettcs:
So far so good, I don't think I have any comments on what's here.

ZoeCMa:
Thanks @brettcs, I do have two questions --

1. For the "passed-through" subcommands (for instance `keep ls`, `ws`, etc.), the "prog" name as displayed in the usage doesn't look right. E.g. the output of `python3 arvcli.py keep ls -h`:
   ```console
   $ python3 arvcli.py keep ls -h
   usage: arvcli.py [-h] [--retries RETRIES] [-s] [--version] locator
   ```
   This should have said `usage: arvcli.py keep ls ...`. I think explicitly adding a `prog=` parameter to the `main()` functions of the passed-through command modules may help.
2. As noted in the comments here https://github.com/arvados/arvados/pull/389/changes#diff-e3f28f197edb4b68f20f91e7749f8533bfd5315611f07566ebd16acb30733464R244-R245 I was using the undocumented attribute `.prog` of the parser instance. I wonder if there's a better way.

brettcs:
I think the requirement is that the usage line look like:

```
usage: arv SUBCOMMAND [options ….]
```

`arv` is the main way we expect users to run the tool. That's what we'll document. So that's what we should show. If users invoke it a different way like `python3 -m arvados.commands.arvcli`, that's fine but they're doing something advanced and we don't need to cater to them.

So I think you have two options. One is, like you said, explicitly hardcoding `prog` with the subcommand for different subparsers. So `prog='arv create'`, `prog='arv keep'`, etc.

The second option is to just not use `prog` at all, and hardcode the command directly into `usage` and other messages.

I think either is fine. Do whatever's easier for you to implement. I think these basic rationales apply to both your questions.

ZoeCMa:
I think there's another, arguably simpler way: just hard-code `prog="arv ..."` into the CLI parsers of the individual standalone commands, e.g.

* `prog="arv keep docker"` in `keepdocker.py`;
* `prog="arv ws"` in `ws.py`;

etc. This may cause some discrepancy if the scripts are invoked as `arv-keepdocker`, `arv-ws` etc., but am I right in saying that we plan to eliminate these aliases in the longer term?

brettcs:
I would personally be fine with that. I had another idea that is also simple and has some 
additional benefits: when we dispatch, before we call the main function for another tool, set `sys.argv[0]` to match the user's invocation. e.g.,

```diff
diff --git i/sdk/python/arvados/commands/arvcli.py w/sdk/python/arvados/commands/arvcli.py
index 157b228090..2379a7cc5c 100644
--- i/sdk/python/arvados/commands/arvcli.py
+++ w/sdk/python/arvados/commands/arvcli.py
@@ -832,10 +832,10 @@ def dispatch(arguments=None):
     method = getattr(args, "method", "")
 
     # Are we calling an external command?
-    ext_module = cmd_parser.external_command_modules.get(
-        f"{args.subcommand} {method}" if method else args.subcommand
-    )
+    ext_cmd = f"{args.subcommand} {method}" if method else args.subcommand
+    ext_module = cmd_parser.external_command_modules.get(ext_cmd)
     if ext_module is not None:
+        sys.argv[0] = f"arv {ext_cmd}"
         _handle_external_command(ext_module, remaining_args)  # Exits.
 
     # Are we doing an API resource call?
```

Since this is where argparse gets the default `prog`, it would automatically update the help message. It would also be reflected in process lists, which is a nice user affordance.

We are probably *not* going to remove the "old" scripts like `arv-keepdocker` and `arv-ws`. Those are very cheap to maintain and removing them would require users to update a lot of scripts and automation for no real benefit. It's not a good trade-off.

ZoeCMa:
Thanks, I tried the `sys.argv[0]` mutation and it works for manual testing. But in the PyTest script I had to use extra caution:

https://github.com/arvados/arvados/commit/b0eecece5936e4a8282d8c718d1e989efef608bf#diff-396bb5c75914466d830e4067be725d9f08d02af4eba695abf2e845d5146b321fR110 (Line 110 in `tests/test_arvcli.py` in case the anchor doesn't work)

Without this detour via `sys.modules`, the next test `test_passthrough_command_usage_prog_name()` fails for the command modules `arvados.commands.keepdocker`, `arvados.commands.get`, and `arvados.commands.put`. These three modules initializes their argparsers at import-time, in the top-level module bodies, so the side effects that depend on `argv[0]` (via their top-level argparser-initialization code) seems rather unpleasant in PyTest.

I wonder if it would be wise to revise those three modules so at least their argparser init could happen inside a function and not at import-time?

ZoeCMa:
This draft PR is superseded by the branch `401-arvcli-help-message-improvements` where the commit messages follow our coding standards: https://github.com/arvados/arvados/compare/main...401-arvcli-help-message-improvements

A PR is in preparation.

ZoeCMa:
Let us continue our discussions about issue at #401. Thank you!

brettcs:
> I wonder if it would be wise to revise those three modules so at least their argparser init could happen inside a function and not at import-time?

This sounds good to me.

ZoeCMa:
Superseded by #414

PR #406, opened 2026-05-29
392: `arv edit` subcommand

### LINKED ISSUE #392: Add `arv edit` subcommand to arvcli
`arv edit` works the same as `arv create` with the following differences:

* Instead of a resource type, the user must pass an Arvados object UUID. This may be followed by a list of field names on that object like `name`, `group_class`, etc. The argument to get the list of field names can use `nargs=argparse.ZERO_OR_MORE`.
* The initial object to edit is retrieved from the API using `RESOURCE/get` with the UUID. If a list of field names was given, it is passed to this call as the `select` argument.
    Hint: The resources in the discovery document have a `uuidPrefix` field that tells you the middle part of the Arvados UUID for those objects. You can use this information to figure out what resource corresponds to the user's UUID.
* After the user finishes editing, the object should be updated using `RESOURCE/update` with the UUID, and the user's input as the body.

Other than this, the flow is the same: we use the same logic for starting the editor, for prompting the user about input errors, etc.

ZoeCMa:
392-arv-edit-subcommand @ 9a4de953e674eb222e68bcc7e18872625d6cdbe1

https://ci.arvados.org/job/developer-run-tests/5088/

  * One notable difference from #392 is how we select the fields of the initial object to be edited. In the issue:
    > If a list of field names was given, it is passed to this call as the select argument.

    Currently I'm instead doing field selection by pulling the full object and doing the filtering _on the client side_ (the same as how the old Ruby `arv` client does it). 
    The reason is easier error handling on the client side: an invalid field in the `select` argument will cause an API error, and we'd have to handle that, but on the client side we can simply ignore invalid fields. 
  * Field selection is made case-insensitive.
  * If no fields end up selected, we don't present the user with an empty object (which looks bewildering). We show the unmodified full object instead.
  * Tmpfile name now includes the readable resource name in addition to the UUID; see c842e2781975065a497b2b3e32020310a4fabaed
  * N/A
  * Automated testing: yes
  * Manual testing involves running the `python3 arvcli.py edit` command with various `fields` positional parameters against a dev build of the API server, and observe the behavior and UX.
  * Yes
  * N/A
  * N/A
  * N/A
  * The possible editable object types are dynamically generated from the server-provided discovery document.
  * Error messaging based on API server error output is generic and minimally dependent on message details.
  * Yes for new Python code

Closes #392.


sdk/python/arvados/commands/arvcli.py
```diff
@@ -642,6 +672,51 @@ def add_resource_subcommands(self):
...
+
+        self.uuid_type_map = _ArgUtil.make_uuid_to_resource_map(
+            self.discovery_document.get("schemas", {})
+        )
+
+        edit_parser = self.subparsers.add_parser(
+            "edit", help="Edit Arvados object using external editor"
+        )
+        edit_parser.add_argument(
+            "uuid", help="UUID of the object to be edited",
```
brettcs: This surprised me later in the code when I saw `uuid[1]`. Usually in our codebase a `uuid` is a plain string, but I knew grabbing the second character out of it made no sense.

At minimum, I think you should set a `dest` for this argument so the internal name makes clearer that this is a more detailed object.

It would be even nicer if, instead of returning a plain tuple, we returned a dataclass with named attributes for the UUID string and resource type. That would be easier to understand at the site of use than a numeric index. `generic_uuid` could become a classmethod constructor like `parse` on the dataclass.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -108,6 +135,15 @@ def parameter_key_to_argument_name(parameter_key: str) -> str:
         """
         return "--" + parameter_key.replace("_", "-")
 
+    @staticmethod
+    def camel_case_to_snake(text: str) -> str:
```
brettcs, line 152: I wasn't gonna make a comment, but since you have a test case for the limitation, consider this implementation:

```python
def camel_case_to_snake(text: str) -> str:
    s = re.sub(r'([^A-Z])([A-Z])', r'\1_\2', text)
    s = re.sub(r'(.)([A-Z])([^A-Z])', r'\1_\2\3', s)
    return s.lower()
```

brettcs:
> Currently I'm instead doing field selection by pulling the full object and doing the filtering _on the client side_ (the same as how the old Ruby `arv` client does it).
> The reason is easier error handling on the client side: an invalid field in the `select` argument will cause an API error, and we'd have to handle that, but on the client side we can simply ignore invalid fields.

I am fine with handling things client-side, and I'm fine handling them case-insensitive. But if the user specifies an invalid field name, we should report that as an argument error and exit. Silently ignoring the fields has the potential to confuse the user if they typo a field name and then that field doesn't appear in their editor.

ZoeCMa:
> But if the user specifies an invalid field name, we should report that as an argument error and exit. Silently ignoring the fields has the potential to confuse the user if they typo a field name and then that field doesn't appear in their editor.

I think this is a very reasonable concern. Even a warning message to the terminal is likely to get wiped by the editor, when the user works in the TUI rather than GUI. I'm going to do the exit thing.

sdk/python/arvados/commands/arvcli.py
ZoeCMa: Thanks, this is a very important point. I just pushed e33cc52d80b7d3952ad42f4f7ac3613ad43c4d85 in which a dataclass is used to hold the "parsed UUID" and the arg dest is renamed `uuid_info` (validated original UUID is now `uuid_info.uuid` and the resource name is `uuid_info.resource_type`).

It turns out that the positional argument doesn't play well with `dest=` kwarg (see python/cpython#117834) so I'm instead doing

```python
edit_parser.add_argument("uuid_info", metavar="UUID")
```
So the usage still says `usage: arvcli.py edit UUID [fields ...]`

sdk/python/arvados/commands/arvcli.py
ZoeCMa: I guess this depends on how defensive we want to be. It seems our naming convention in the discovery doc mostly follows [this guide](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) ([referenced](https://google.aip.dev/190#:~:text=Unless%20otherwise%20specified%2C%20definitions%20must%20use%20UpperCamelCase%20names) in Google's "AIP" docs [like RFCs for Google API]). This requires "hard" CamelCasing even for abbreviations, like "Api". If we follow this convention, we can get away with the original one-liner.

But if we want to be more defensive, I wrote a small(ish) function with comments --

```python
    @staticmethod
    def camel_case_to_snake(text: str) -> str:
        """Simple converter of CamelCase text to so-called 'snake_case' (lower
        case with underscore).
        """
        output = []
        upper = [char.isupper() for char in text]
        # Skip non-letter characters in "edge-detection"; see below.
        skip = [not char.isalpha() for char in text]
        # Locate "rising" (aB -> a|B) and "falling" (ABCXy -> ABC|Xy) edges.
        for i, char in enumerate(text):
            if i > 0 and upper[i]:  # Only upper-case letter can be an edge.
                rise = not (upper[i - 1] or skip[i - 1])
                fall = (
                    upper[i - 1] and i + 1 < len(text)
                    and not (upper[i + 1] or skip[i + 1])
                )
                if rise or fall:
                    output.append("_")  # Insert underscore at edge.
            output.append(char.lower())
        return "".join(output)
```
which passes for these test cases (i.e. it handles more cases than the regex-sub suggestion, and it doesn't use regex):
```python
@pytest.mark.parametrize("text,expected", (
    ("", ""),
    ("a", "a"),
    ("A", "a"),
    ("A@B", "a@b"),
    ("A._b", "a._b"),
    ("AList", "a_list"),                                                        
    ("ALL_CAPS", "all_caps"),
    ("snake_case", "snake_case"),
    ("mixedCase", "mixed_case"),
    ("HTTPClient", "http_client"),
    ("HTTP_Client", "http_client"),
    ("getURL", "get_url"),
    ("api2CLI", "api2cli"),
    # Cases above are defensive; the one below is illustrative.
    ("ApiClientAuthorization", "api_client_authorization")
))
def test_camel_to_snake(text, expected):
    assert arvcli._ArgUtil.camel_case_to_snake(text) == expected
```
I kinda like this function and I feel it _may_ be easier to follow than regex, but I've no special attachment to it. Shall we keep it?

brettcs: It might make the most sense to go up a level and talk about testing philosophy.

It's good to test a variety of expected cases, especially corner cases. However, it's also good to limit yourself to testing inputs you actually expect. Doing otherwise spends time on tests for no benefit, and can lead future readers to believe a particular behavior is required when it isn't, actually.

When your function is dealing directly with user input, the input could be basically anything and so there are very few limits on testing. But if you know there are limits on input—as here, where the function only deals with "resource names in the Arvados discovery document"—there's no benefit to testing outside those limits.

So maybe the better change here is: don't change the implementation at all, edit the test cases to be a sample of Arvados resource names (taking care to choose any particularly tricky ones), and don't worry about inputs that don't exist like `HTTPClient`.

ZoeCMa: Great points. Also we've been doing pretty well in keeping "tricky ones" away from our schema definitions :)

(In cde6f216e36e0f4e06aa86cea76ae2eaabe6a64c the tests are revised.)

ZoeCMa:
> > But if the user specifies an invalid field name, we should report that as an argument error and exit. Silently ignoring the fields has the potential to confuse the user if they typo a field name and then that field doesn't appear in their editor.
> 
> I think this is a very reasonable concern. Even a warning message to the terminal is likely to get wiped by the editor, when the user works in the TUI rather than GUI. I'm going to do the exit thing.

This is done in 14acf2c5013d086afc2751665e2ec6a4ae697296.

If it looks ready, I'm going to make the rebase and resolve conflicts.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -45,12 +47,49 @@ def group_uuid(text: str) -> str:
...
         raise argparse.ArgumentTypeError(
             f"Invalid UUID for Arvados project or group: {text}"
         )
 
+    @dataclass(frozen=True)
+    class UUIDInfo:
+        """'Interpreted' Arvados UUID object with resource type info."""
+        uuid: str
+        resource_type: str  # value in CamelCase
+        resource_type_alt: str  # value in snake_case
```
brettcs: Please consider something more descriptive than "alt" for the name. "lower" or "snake" would be fine. It seems like it would also be a good trade to abbreviate the first part to have a longer description, like `rtype_snake_case` if you want.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -731,19 +825,93 @@ def _handle_resource_method(api_client, resource, args) -> NoReturn:
     sys.exit(_call_resource_method(arv_method, method_args, args.format))
 
 
+def _select_fields(
+    src: Mapping[str, Any], fields: Iterable[str]
+) -> Mapping[str, Any]:
```
brettcs, line 887: This is more code than you need. Consider instead:

```python
return {k: src[k] for k in fields} or src
```

sdk/python/tests/test_arvcli.py
```diff
@@ -1066,3 +1176,80 @@ def test_json_input_not_an_object(self, setup_editor, run_arvcli):
...
+        assert exit_code == 2
+
+    @pytest.mark.usefixtures("reset_test_server_db")
+    def test_edit_collection_name(
+            self, simple_api_client, setup_editor, run_arvcli
+    ):
+        uuid = run_test_server.fixture("collections")[
+            "collection_owned_by_active"
+        ]["uuid"]
+        old_obj = simple_api_client.collections().get(uuid=uuid).execute()
```
brettcs: Please use the entire fixture rather than fetching it from the API server. An API call makes the test needlessly slow.

sdk/python/tests/test_arvcli.py
```diff
@@ -1066,3 +1176,80 @@ def test_json_input_not_an_object(self, setup_editor, run_arvcli):
...
+        exit_code, out, err = run_arvcli(["edit", uuid])
+
+        assert exit_code == 0
+        assert not err
+        new_obj = json.loads(out)
+        assert new_obj["uuid"] == uuid
+        assert new_obj["name"] == edited_obj["name"]
+        assert ciso8601.parse_datetime(new_obj["modified_at"]) >= timestamp
+
+    def test_edit_collection_did_not_edit(self, setup_editor, run_arvcli):
```
brettcs: Please add another test case that does this with fields selected.

brettcs:
> If it looks ready, I'm going to make the rebase and resolve conflicts.

I made a few comments but they're all pretty small, please both rebase and address them. I'm not worried about losing context. Thanks.

sdk/python/tests/test_arvcli.py
ZoeCMa: Fixed in 767e404746eb48013dbec667dbac0bfeb9358940.

ZoeCMa:
I did a rebase. The tests are being run: https://ci.arvados.org/job/developer-run-tests/5096/

PR #409, opened 2026-06-03
408: Fix `arvcli config get` help message showing spurious parameters

### LINKED ISSUE #408: Python CLI client: `config get -h` subcommand outputs spurious parameters in help message.
First of all, although `arvcli.py` is not currently exposed, we can observe this problem by manually running the script. In the console output below, all of the parameters except `-h/--help` are spurious.

```console
$ python3 arvcli.py config get -h
usage: arvcli.py config get [-h] [-a STR] [-f STR] [-k STR] [-o STR] [-u STR]
                            [-t STR] [-s STR] [-p STR]

options:  
  -h, --help            show this help message and exit
  -a STR, --alt STR     Data format for the response. Default: json.
  -f STR, --fields STR  Selector specifying which fields to include in a
                        partial response.
  -k STR, --key STR     API key. Your API key identifies your project and
                        provides you with API access, quota, and reports.
                        Required unless you provide an OAuth 2.0 token.
  -o STR, --oauth-token STR
                        OAuth 2.0 token for the current user.
  -u STR, --userip STR
  -t STR, --trace STR
  -s STR, --strict STR
  -p STR, --pp STR

```

This is caused by a combination of the following actions --

1. In `arvcli.py` we initialize an API client object, then retrieve the equivalent of the discovery document from the API client object's resource description dict `_resourceDesc`/`_rootDesc`.
2. As a side effect, `googleapiclient` injects top-level parameters and (not well-documented) ["stack query parameters"](https://github.com/googleapis/google-api-python-client/blob/6e471c075039dfef24e28d11658e03d5c949c7c3/googleapiclient/discovery.py#L843-L845) in to the resource description dict, even if the Arvados stack doesn't use those parameters, unlike Google. This happens when we make a `googleapiclient.discovery.Resourc` object.
3. Even before we do any work in `arvcli.py`, the initialization of the API client object causes one Resource object to be made, and it's the `configs` resource; long story short, a [`KeepClient` object](https://github.com/arvados/arvados/blob/fa9b3106ebe873422fb34ca79ada65a11f4f9c5f/sdk/python/arvados/api.py#L305) retrieves the [default storage classes](https://github.com/arvados/arvados/blob/fa9b3106ebe873422fb34ca79ada65a11f4f9c5f/sdk/python/arvados/keep.py#L928) from the cluster's public [`configs` resource](https://github.com/arvados/arvados/blob/fa9b3106ebe873422fb34ca79ada65a11f4f9c5f/sdk/python/arvados/util.py#L335).

Overall this means the `configs` resource come pre-populated with the injected parameters.

To work around this for now, we may ignore-list the injected parameters when we convert the resource description dict into CLI parameters. The injected parameters can be programmatically obtained by combining the [top-level parameters from the discovery doc](https://github.com/arvados/arvados/blob/fa9b3106ebe873422fb34ca79ada65a11f4f9c5f/sdk/python/arvados-v1-discovery.json#L22-L50) with those [hard-coded in `googleapiclient.discovery`](https://github.com/googleapis/google-api-python-client/blob/6e471c075039dfef24e28d11658e03d5c949c7c3/googleapiclient/discovery.py#L130).

ZoeCMa:
`408-fix-arvcli-config-get-spurious-parameters` @ d9c5a1b314681f8f8350e39e30325feacdab1a1d

https://ci.arvados.org/job/developer-run-tests/5091/ (crunchrun intermittent failure affecting the tests but is unrelated)

Fixes #408.

  * Implemented the "parameter ignore-list" idea suggested by myself.
  * N/A
  * Yes for auto testing
  * Manual testing was done to ensure the output of `-h` with various resource-method combos look alright (notably `config get`).
  * Yes
  * N/A
  * N/A
  * N/A
  * At the layer of the CLI client, there should be little to worry about client/server compatibility, as we still follow the principle of discovery-based API.
  * A related issue is the compatibility of the "official" Google-API client with the Arvados "Google-esque" API. Behavior of the `googleapiclient` package may include Google-specific quirks and it'll be worthwhile to explore how to minimize our exposure.
  * Yes

### DIFF
--- a/sdk/python/arvados/commands/arvcli.py
+++ b/sdk/python/arvados/commands/arvcli.py
@@ -18,6 +18,7 @@ The `ArvCLIArgumentParser` class, specializing the standard Python
 
 import abc
 import argparse
+from collections.abc import Container, Iterable, Mapping
 from contextlib import AbstractContextManager
 import functools
 import importlib
@@ -29,8 +30,9 @@ import shutil
 import subprocess
 import sys
 from tempfile import NamedTemporaryFile
-from typing import Any, Mapping, NoReturn, TextIO
+from typing import Any, NoReturn, Optional, TextIO
 import arvados
+from arvados.api import apiclient_discovery
 import arvados.commands._util as cmd_util
 from ruamel.yaml import YAML, YAMLError
 yaml = YAML(typ="safe", pure=True)
@@ -109,7 +111,10 @@ class _ArgUtil:
         return "--" + parameter_key.replace("_", "-")
 
     @staticmethod
-    def get_method_options(method_schema):
+    def get_method_options(
+        method_schema: Mapping[str, Any],
+        ignored_parameters: Optional[Container[str]] = None
+    ):
         """Generate command-line options, in the form of "-f/--foo", from the
         parameters as defined by the API method schema in the discovery
         document.
@@ -142,8 +147,10 @@ class _ArgUtil:
 
         Arguments:
 
-        * method_schema: dict --- Dict object from the parsed discover document
-          that defines a method.
+        * method_schema: Mapping[str, Any] --- Dict object from the parsed
+          discover document that defines a method.
+        * ignored_parameters: Optional[Container[str]] --- If provided, the
+          parameters that are in `ignored_parameters` will not be processed.
         """
         parameters_schema = method_schema.get("parameters", {}).copy()
         # If the method comes with the "request" field, add another parameter
@@ -162,6 +169,8 @@ class _ArgUtil:
                 }
         argument_key_abbrevs = set("h")  # prevent conflict with "help"
         for parameter_key, parameter_dict in parameters_schema.items():
+            if ignored_parameters and parameter_key in ignored_parameters:
+                continue
             parameter_kwargs = {
                 "required": parameter_dict.get("required", False)
             }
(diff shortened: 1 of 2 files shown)


sdk/python/arvados/commands/arvcli.py
```diff
@@ -29,8 +30,9 @@
 import subprocess
 import sys
 from tempfile import NamedTemporaryFile
-from typing import Any, Mapping, NoReturn, TextIO
+from typing import Any, NoReturn, Optional, TextIO
 import arvados
+from arvados.api import apiclient_discovery
```
ZoeCMa: Actually not sure if I should do this (importing from "our" package) or just import the `googleapiclient` 3rd-party package directly.

brettcs: Please prefer a direct import. Removing a layer of indirection is easier for the reader to follow.

sdk/python/arvados/commands/arvcli.py
```diff
@@ -109,7 +111,10 @@ def parameter_key_to_argument_name(parameter_key: str) -> str:
         return "--" + parameter_key.replace("_", "-")
 
     @staticmethod
-    def get_method_options(method_schema):
+    def get_method_options(
+        method_schema: Mapping[str, Any],
+        ignored_parameters: Optional[Container[str]] = None
```
brettcs: Not a big deal but you can simplify this to:

```python
ignored_parameters: Container[str] = (),
```

That provides the same behavior without any need to handle the `None` case.

ZoeCMa:
Thanks for the suggestions. Done in 72eecc3d969d234edfd9ef28bf642816886d5884.

brettcs:
The LGTM, thanks. It looks like there's some overlap between this branch and #406. What's the best way I can handle this that makes life easier for you? Should I merge this now, wait until after #406 is merged, something else? Please let me know.

ZoeCMa:
Thanks, I think it's okay to merge into main now.

I think it's mostly likely the best use of our time if I resolve the conflicts in outstanding PR branches by myself.

I can do this once those branches are substantially ready to merge; then I'll do a rebase to the new main, and fix the conflicts.

Then, depending on whether I can force-push to the `git.arvados.org` origin (not sure but probably I can't) --

- If I can, I'd force-push to the "true origin", letting arvados-bot do the syncing here.
- If I can't, I'd push new clean branches, make new PRs with notes saying they supersede the old ones, and close the old ones.

How about this?

brettcs:
You can't force push (nobody can) but you can delete the original branch on the server, then push a new one with the same name. It's basically the same thing with more steps.

PR #411, opened 2026-06-03
410: Add confirmation dialog before trash and restore operations

### LINKED ISSUE #410: Add confirmation dialog before moving items to trash or restoring from trash
**Problem**
Workbench 2 currently moves collections and projects to trash immediately when the user clicks "Move to Trash" in the context menu or toolbar, with no confirmation step. The same applies to the "Restore" action on trashed items. A misclick can unintentionally trash a resource.

**Proposed Solution**
Add a confirmation dialog that appears before any trash or restore operation is executed.

gilsdom1-roche:
## Description

Adds a confirmation dialog before trash and restore operations in Workbench 2.

Currently, clicking "Move to Trash" immediately trashes the resource with no
confirmation step. This change intercepts all trash/restore entry points and
shows a ConfirmationDialog before proceeding.

Closes #410

## Changes

- **New**: `trash-confirm-dialog.tsx` — confirmation dialog component using
  existing `ConfirmationDialog` + `withDialog` pattern
- **New**: `openTrashConfirmDialog` thunk in `trash-actions.ts` — reads resource
  names from the store, opens dialog with context-appropriate text
- **Modified**: All 5 call sites that previously called `toggleResourceTrashed`
  directly now call `openTrashConfirmDialog`
- **Modified**: `workbench.tsx` — registers `<TrashConfirmDialog />`
- **Modified**: `AUTHORS` — added contributor

## Dialog behavior

- **Single item**: Shows resource name
  (e.g., `Are you sure you want to move "My Collection" to trash?`)
- **Multiple items**: Shows count
  (e.g., `Are you sure you want to move 5 items to trash?`)
- **Restore**: Same pattern with restore-specific text
- Cancel closes the dialog with no side effects

## PR Checklist

* All agreed upon points are implemented / addressed.
  *All 5 trash/restore entry points (context menus, toolbar, trash panel restore button) now show a confirmation dialog.*
* Anything not implemented has a follow-up story.
  *Everything from the issue is implemented. No follow-up needed.*
* Code is tested and passing, both automated and manual.
  * Manually tested on two Workbench instances — single item trash, multi-select trash, restore from trash panel, restore from context menu. No automated tests added, consistent with existing confirmation dialogs (e.g., RemoveProcessDialog).*
  *Branch is based on current main.*
  *The implementation is consistent with the request received from stakeholders. Stakeholders haven't been presented with this solution.*
  *No documentation changed*
* Behaves appropriately at the intended scale.
  *Dialog is a lightweight MUI component with no performance impact. Tested with single and multi-select operations.*
* Considered backwards and forwards compatibility issues.
  *Purely client-side change, no API modifications.*
* Follows our coding standards, including GUI style guidelines.
  *Uses existing ConfirmationDialog component and withDialog/dialogActions pattern, consistent with RemoveProcessDialog and other confirmation dialogs in the codebase.*

## AI Disclosure

This implementation was created with the assistance of AI (Claude, via Ona). The code was reviewed and manually tested by the contributor.


brettcs:
This PR started a little debate within the core team about whether it was the right approach. When people are developing new workflows in Arvados, it's common for them to accumulate data they don't want to keep and would like to clean up. Requiring a modal confirmation for every deletion would annoy these users.

I can understand and agree with the core concern: it is too easy to accidentally trash data without realizing it. One issue that I think contributes to this is our snackbar timing. It's not 100% consistent but generally Workbench disappears success snackbars after 2 seconds, and error snackbars after 4. This is [below the Material guidelines](https://m3.material.io/components/snackbar/accessibility) which suggest 4–10 seconds for most applications.

I would like to at least start to address this by increasing our snackbar timing to 8 seconds on success and 10 seconds on error across the board. This gives users time to check that the confirmation matches what they wanted. It also doesn't require any changes to the logic.

From there, I think another improvement could be to offer an "Undo" action directly from the snackbar, to help streamline the process of correcting mistakes.

gilsdom1-roche:
I agree that increasing the snackbar visibility makes sense. This is something that annoyed me in the past but never enough to mention it.

Regarding the cleanup of workflow run artifacts. First, this change adds one more click, both to single object deletions and multi object deletions. Second, and this may be specific to us, we have a much larger user base that are using Arvados for data management  and some who use it to execute workflows. From them we got the feedback that accidental deletion that goes undetected is a real issue.

Anyway, we may have different opinions on the "how" but at least we agree that something needs to be done and any improvement is appreciated.

PR #417, opened 2026-06-15
374: Recommend pipx installation in documentation

### LINKED ISSUE #374: User guide recommends installing Python tools with pipx
Anywhere under `doc/user` we give the user instructions to install a Python tool, we should offer them three options in this order:

1. Install `pipx` and run `pipx --include-deps arvados-tools`
2. Install the apt repository and corresponding Debian package
3. Install the yum repository and corresponding RHEL package

Most pages should already have steps 2 and 3. But we want pipx to be our primary installation method for users: it works on more distros and doesn't require fussing with system configuration.

We do not need to tell the user the name of the specific package that actually provides the tool they're installing. For the audience of the user guide, that doesn't matter.

We do not need to give the user instructions for how to set up and install in a virtualenv manually. For the audience of the user guide, that's too much technical detail.

We do not want to change pages outside the user guide in this ticket. e.g., the SDK reference _can_ talk about installing `arvados-python-client` in a virtualenv, since people developing their own scripts should be more comfortable with that.

ZoeCMa:
`374-doc-recommend-pipx` @ 2fdaf4c62cf88e4f9d467556819a96ae2e473741

https://ci.arvados.org/job/developer-run-tests/5098/

  * _Yes_
  * _N/A_
  * `test doc` passes ;
  * Manual test: inspecting the built site in a browser for both content and format.
  * _Yes_
  * _N/A_
  * _Yes_, within the scope of #374 
  * _N/A_
  * _N/A_; installation instructions are not concerned with the server.
  * _Yes_

Closes #374.

### DIFF
--- a/doc/_includes/_setup_redhat_repo.liquid
+++ b/doc/_includes/_setup_redhat_repo.liquid
@@ -31,7 +31,7 @@ EOF</span>
 {%- if modules_to_enable != nil %}
 # <span class="userinput">dnf module enable {{ modules_to_enable }}</span>
 {% endif -%}
-{%- if packages_to_install != nil -%}
+{%- if packages_to_install != nil %}
 # <span class="userinput">dnf install {{ packages_to_install }}</span>
 {% endif -%}
 </code></pre>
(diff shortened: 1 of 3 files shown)


brettcs:
I see you also made changes that make the instructions more appropriate for the Python `arv` command. Thanks for doing that, that's helpful.

I pushed a commit with some subjective style tweaks. My main goal was just to simplify the text as much as possible. I removed or deprioritized detail when it didn't feel necessary.

Is this version okay by you? If so I'll merge. Thanks.

ZoeCMa:
> Is this version okay by you? If so I'll merge. Thanks.

Thanks, this looks nice.

Issue #433, opened 2026-06-23
Remove filtering by pipeline and job in Python WebSocket client; improving filter validation


ZoeCMa:
The `arvados#pipelineInstance` and `arvados#job` object kinds have been removed since bf521e2fd9fae16bf5738f988496a3dbb1e32833 and they should be removed from `arvados/commands/ws.py`.

In addition, the `--filters` option should have the input filter value validated by `arvados.commands._util.validate_filters`.

PR #436, opened 2026-06-23
432: Python client commands exit consistently

### LINKED ISSUE #432: Inconsistent behaviors of exiting from main function in Python arv-* scripts
Currently in the Python `arv` command script `arvcli.py`, for external-module commands such as `arv keep get`:

https://github.com/arvados/arvados/blob/17c69aa16e6981eadf9297ec8ca5de1898eb18fc/sdk/python/arvados/commands/arvcli.py#L814-L820

This assumes that the external module's `main()` function returns an integer status suitable to be passed to `sys.exit()`.

However, this assumption doesn't hold. For example, in the `main()` of `put.py`:

https://github.com/arvados/arvados/blob/17c69aa16e6981eadf9297ec8ca5de1898eb18fc/sdk/python/arvados/commands/put.py#L1399-L1400

This will cause `arvcli.py` to exit abnormally, _and_ duplicate the `output` to stderr.

The `arv-*` scripts under `bin/` also reflect the current inconsistency. Some simply calls `main()` while others say `sys.exit(main(sys.argv[1:]))`.

I think we can improve consistency for maintainability, by implementing this convention:

1. The `main()` functions of the "command" modules under `arvados/commands` are `NoReturn` functions; they never returns, but always terminates by calling `sys.exit()` with the appropriate status.
   * The reason is that most (all?) of them calls `argparse.ArgumentParser.parse_args()` and by default, this is already a possible exit point: the program exits with status `2` upon invalid CLI arguments.
2. Command scripts under `bin/`, and the external-module handler in `arvados/commands/arvcli.py`, calls the imported `main()` rather than `sys.exit(main())`, as a result of 1).

ZoeCMa:
`432-python-commands-exit-consistently` @ 3b681097cedd466a0a312e9797dd2da44ac1a990

https://ci.arvados.org/job/developer-run-tests/5104/

  * _Yes_; before the implementation, the `main()` functions behave differently:
    * `arv_copy.py` and `keepdocker.py` call `sys.exit()`, although the former did it a bit incorrectly (calling the top-level `exit()` created by the `site` module for interactive interpreter, rather than `sys.exit()`)
    * `get.py` and `ls.py` return an integer status.
    * `put.py` did sort of both: calling `sys.exit()` on error but returns a string value upon success.
  * After the implementation, all of those `main()` functions are NoReturn (calling `sys.exit()`)
    * For `put.py`, the function that does the majority of work is needed by `keepdocker.py`. This function is now named `do_put()`, which returns a tuple `(integer_code, str_output)`. `do_put()` may still call `sys.exit()` if given nonsensical arguments (because it calls `argparse.ArgumentParser.parse_args()`).
  * See #433 (not related)
  * Tests routines are revised to adapt to the new convention implemented in this change. Manually tested with running `arv {copy|keep ls|keep get|keep put}` commands via arvcli.py.
  * _Yes_
  * _N/A_
  * _N/A_ (largely internal changes without outside observable behavior change)
  * _N/A_
  * _Yes_ (not affected)
  * _Yes_ for new Python code.

Additionally, in `put.py`, it used to be the case that most errors cause exit with status 1. I've made it exit with 2 when the error is caused by incorrect command-line input, and reserved 1 for run-time errors.

Closes #432.

### DIFF
--- a/sdk/python/arvados/commands/put.py
+++ b/sdk/python/arvados/commands/put.py
@@ -1143,8 +1143,8 @@ def desired_project_uuid(api_client, project_uuid, num_retries):
         raise ValueError("Not a valid project UUID: {}".format(project_uuid))
     return query.execute(num_retries=num_retries)['uuid']
 
-def main(arguments=None, stdout=sys.stdout, stderr=sys.stderr,
-         install_sig_handlers=True):
+def do_put(arguments=None, stdout=sys.stdout, stderr=sys.stderr,
+           install_sig_handlers=True) -> tuple[int, str | None]:
     global api_client
 
     args = parse_arguments(arguments)
@@ -1153,7 +1153,6 @@ def main(arguments=None, stdout=sys.stdout, stderr=sys.stderr,
         logger.setLevel(logging.WARNING)
     else:
         logger.setLevel(logging.INFO)
-    status = 0
 
     request_id = arvados.util.new_request_id()
 
@@ -1173,7 +1172,7 @@ def main(arguments=None, stdout=sys.stdout, stderr=sys.stderr,
         # make sure the user provides a complete YYYY-MM-DD date.
         if not re.match(r'^\d{4}(?P<dash>-?)\d{2}?(?P=dash)\d{2}', args.trash_at):
             logger.error("--trash-at argument format invalid, use --help to see examples.")
-            sys.exit(1)
+            return 2, None
         # Check if no time information was provided. In that case, assume end-of-day.
         if re.match(r'^\d{4}(?P<dash>-?)\d{2}?(?P=dash)\d{2}$', args.trash_at):
             args.trash_at += 'T23:59:59'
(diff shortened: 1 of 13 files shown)


sdk/python/arvados/commands/put.py
```diff
@@ -1316,17 +1315,17 @@ def main(arguments=None, stdout=sys.stdout, stderr=sys.stderr,
...
+        return 1, None
     except (CollectionUpdateError, PathDoesNotExistError) as error:
         logger.error("\n".join([
             "arv-put: %s" % str(error)]))
-        sys.exit(1)
+        return 1, None
     except ArvPutUploadIsPending:
         # Dry run check successful, return proper exit code.
-        sys.exit(2)
+        return 1, None
```
ZoeCMa, line 1325: I _think_ `1` would have been the correct code here. This part only concerns dry-runs and I'm not entirely sure though.

brettcs (approved):

PR #442, opened 2026-07-02
440: Controller sends HTTP 400 upon invalid manifest for replace_files.

### LINKED ISSUE #440: When using `replace_files` to create/update collection with invalid manifest text in request body, controller returns 500 instead of 4xx
```python
import json

import arvados


# Invalid manifest: no newline at the end.
bad_manifest = ". d41d8cd98f00b204e9800998ecf8427e+0 0:0:foo"
api_client = arvados.api("v1", num_retries=0)
method = api_client.collections().create(
    replace_files={
        "/new.txt": "manifest_text/foo"
    },
    body={
        "collection": {
            "manifest_text": bad_manifest
        }
    }
)
result = method.execute()
print(json.dumps(result, indent=1))
```

Output (excerpt):

    arvados.errors.ApiError: <HttpError 500 when requesting https://xsat1.halley-mirzam.ts.net:8443/arvados/v1/collections?replace_files=%7B%22%2Fnew.txt%22%3A+%22manifest_text%2Ffoo%22%7D&alt=json returned "line 1: no trailing newline">

Here `500 Internal Server Error` is not an appropriate HTTP status, because this is a client input error.

Moreover, the Python SDK API client built around `googleapiclient`, [will retry upon HTTP 500](https://github.com/googleapis/google-api-python-client/blob/bf14134647a9c5c9e7430b0c5034afbcec89a203/googleapiclient/http.py#L92-L94). With [exponential backoff](https://github.com/googleapis/google-api-python-client/blob/bf14134647a9c5c9e7430b0c5034afbcec89a203/googleapiclient/http.py#L175-L176) and default number of retries == 10, the sleep time can get quickly out of hand.

The client should not retry with the same invalid request and expect a different result, and we need to let the client know this.  The existing implementation in `lib/controller/localdb/collection.go` returns `400 Bad Request` upon invalid input in many other cases, and I think we could do the same here.

Furthermore, the error message from the server is rather ambiguous. It's unclear _whose_ "line 1" it refers to (answer: the value of the `"manifest_text"` key in the JSON POST body), and the reader might think it was referring to the query parameter `replace_files`. We need to say that it's the manifest text that's invalid.

ZoeCMa:
`440-return-http-400-bad-request-on-invalid-manifest-for-replace_files` @ cdd2a326b0612ca0bc7c86edc8f74a01d27c8642

https://ci.arvados.org/job/developer-run-tests/5109/

  * I tried to implement minimally invasive changes to the SDK code: whereas formerly the private method `loadManifest()` returns a plain error created by `fmt.Errorf()`, here it returns one, also created by `fmt.Errorf()`, that wraps the custom sentinel error `ErrInvalidManifestText`. On the controller side, upon calling `Collection.FileSystem()` and checking the error with `errors.Is()`, send the error response with `httpserver.Errorf(http.StatusBadRequest, ...)`.
  * _N/A_
  * I added some test cases on both sides and they pass.
  * _Yes_
  * _N/A_
  * _N/A_
  * The update should minimally affect the most common case (the provided `manifest_text` being actually valid). Upon invalid `manifest_text`, by informing a semantics-compliant client with the 400 code, we may actually prevent inadvertent (and futile) retries.
  * This should improve the interoperability with the `googleapiclient`-based Python client marginally (that is, upon the specific kind of invalid input).
  * Yes (gofmt).

Closes #440.

### DIFF
--- a/lib/controller/localdb/collection.go
+++ b/lib/controller/localdb/collection.go
@@ -290,7 +290,7 @@ func (conn *Conn) applyReplaceFilesOption(ctx context.Context, fromUUID string, 
 			srccoll := &arvados.Collection{ManifestText: providedManifestText}
 			srcfs, err = srccoll.FileSystem(&arvados.StubClient{}, &arvados.StubClient{})
 			if err != nil {
-				if errors.Is(err, arvados.ErrInvalidManifestText) {
+				if errors.Is(err, arvados.ErrMalformedManifestText) {
 					return nil, httpserver.Errorf(http.StatusBadRequest, "replace_files: %w", err)
 				}
 				return nil, err
--- a/lib/controller/localdb/collection.go
+++ b/lib/controller/localdb/collection.go
@@ -289,10 +289,9 @@ func (conn *Conn) applyReplaceFilesOption(ctx context.Context, fromUUID string, 
 			srcfs = nil
 			srccoll := &arvados.Collection{ManifestText: providedManifestText}
 			srcfs, err = srccoll.FileSystem(&arvados.StubClient{}, &arvados.StubClient{})
-			if err != nil {
-				if errors.Is(err, arvados.ErrMalformedManifestText) {
-					return nil, httpserver.Errorf(http.StatusBadRequest, "replace_files: %w", err)
-				}
+			if errors.Is(err, arvados.ErrMalformedManifestText) {
+				return nil, httpserver.Errorf(http.StatusBadRequest, "replace_files: %w", err)
+			} else if err != nil {
 				return nil, err
 			}
 			srcidloaded = srcid
--- a/lib/controller/localdb/collection.go
+++ b/lib/controller/localdb/collection.go
@@ -6,6 +6,7 @@ package localdb
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -289,6 +290,9 @@ func (conn *Conn) applyReplaceFilesOption(ctx context.Context, fromUUID string, 
 			srccoll := &arvados.Collection{ManifestText: providedManifestText}
 			srcfs, err = srccoll.FileSystem(&arvados.StubClient{}, &arvados.StubClient{})
 			if err != nil {
+				if errors.Is(err, arvados.ErrInvalidManifestText) {
+					return nil, httpserver.Errorf(http.StatusBadRequest, "replace_files: %w", err)
+				}
 				return nil, err
 			}
 			srcidloaded = srcid
--- a/sdk/go/arvados/fs_collection.go
+++ b/sdk/go/arvados/fs_collection.go
@@ -1777,16 +1777,16 @@ func splitToToks(src []byte, c rune, toks [][]byte) int {
 	return 3
 }
 
-var ErrInvalidManifestText = errors.New("invalid manifest_text")
+var ErrMalformedManifestText = errors.New("malformed manifest_text")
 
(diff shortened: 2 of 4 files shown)


sdk/go/arvados/fs_collection.go
```diff
@@ -1777,10 +1777,16 @@ func splitToToks(src []byte, c rune, toks [][]byte) int {
 	return 3
 }
 
+var ErrInvalidManifestText = errors.New("invalid manifest_text")
```
brettcs: I propose the error say "malformed" instead of "invalid" (and the name reflect that). I think that's just a little bit clearer.

sdk/go/arvados/fs_collection.go
```diff
@@ -1777,10 +1777,16 @@ func splitToToks(src []byte, c rune, toks [][]byte) int {
...
+
+func invmtErrorf(format string, args ...interface{}) error {
```
brettcs: Consider instead the name `badmtxtErrorf`. I think the `inv` abbreviation does not spring immediately to mind, and `mtxt` is a commonly used abbreviation throughout our Go code.

sdk/go/arvados/fs_collection.go
```diff
@@ -1777,10 +1777,16 @@ func splitToToks(src []byte, c rune, toks [][]byte) int {
...
+	return fmt.Errorf("%w: %s", ErrInvalidManifestText, fmt.Sprintf(format, args...))
```
brettcs: Rather than nesting format calls, consider instead:

```go
return fmt.Errorf("%w: "+format, ErrMalformedManifestText, args...)
```

lib/controller/localdb/collection.go
```diff
@@ -289,6 +290,9 @@ func (conn *Conn) applyReplaceFilesOption(ctx context.Context, fromUUID string,
 			srccoll := &arvados.Collection{ManifestText: providedManifestText}
 			srcfs, err = srccoll.FileSystem(&arvados.StubClient{}, &arvados.StubClient{})
 			if err != nil {
```
brettcs: Please avoid pyramiding by doing all error checks at the top level like so:

```go
if errors.Is(err, arvados.ErrInvalidManifestText) {
    …
} else if err != nil {
    return nil, err
}
```

ZoeCMa: Fixed in aa3011ed8059846c2dd96081673dd9552b1be9d2.

sdk/go/arvados/fs_collection.go
ZoeCMa: Fixed in be99f229017205a0172a9e889a1f13e641fab713.

sdk/go/arvados/fs_collection.go
ZoeCMa: Fixed in be99f229017205a0172a9e889a1f13e641fab713.

sdk/go/arvados/fs_collection.go
ZoeCMa: Hmm, a call to `fmt.Errorf()` like this would require creating a new slice with `ErrMalformedManifestText` at its head; otherwise we get `too many arguments in call to fmt.Errorf` error.  So in aa3011ed8059846c2dd96081673dd9552b1be9d2 I did just that. Does that look good?

ZoeCMa: (modified again in ebf5628b5d72b229811a587cc8101829301e680d to use `slices.Concat` which was added in go1.22)

PR #449, opened 2026-07-10
Reorganize cluster activity report to improve maintainability

brettcs:
426-cluster-activity-rework @ a96bca90691f1cc3ba3559d4e8b3df80cd4eec6a

<https://ci.arvados.org/job/developer-run-tests/5114/>

This branch fixes a couple minor bugs in the cluster activity report, but more importantly, it reorganizes the code to lay the groundwork for the various improvement issues that have been filed. For more detail, see the commit message for 733f9b5d80ee37b14fd20bf15bee6da984a02ab9 specifically.

  * This branch does not close any issue, but just makes it easier to close outstanding issues. Because of that: yes, there are absolutely still bugs in this that I did not address. I'm already a little concerned this branch is too big. I specifically left some bugs and quirks as-is to make sure that I wasn't making any changes to the output of the tool that I couldn't explain. Any bugs that aren't strictly new to this branch I will probably ask to spin off into a separate issue.
  * N/A
  * See above. In addition, the code has generated real reports on two production clusters.
  * No recent changes in cluster activity report
  * N/A
  * N/A (the only user-facing changes are minor improvements in the accuracy of output data)
  * This probably does use a little more RAM while building the report, but I expect the additional overhead is swamped by the amount of raw data that needs to just be loaded from Arvados.
  * N/A
  * I hope so, I did my best.

### DIFF
--- a/tools/cluster-activity/arvados_cluster_activity/report.py
+++ b/tools/cluster-activity/arvados_cluster_activity/report.py
@@ -44,9 +44,7 @@ class BytesFormatter:
 
 
 bytes_base2_fmt = BytesFormatter(1024, ['KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB'])
-format_with_suffix_base2 = bytes_base2_fmt
 bytes_si_fmt = BytesFormatter(1000, ['KB', 'MB', 'GB', 'TB', 'PB', 'EB'])
-format_with_suffix_base10 = bytes_si_fmt
 
 def cost_fmt(cost):
     return f'${cost:,.02f}'
--- a/tools/cluster-activity/arvados_cluster_activity/report.py
+++ b/tools/cluster-activity/arvados_cluster_activity/report.py
@@ -3,22 +3,23 @@
 #
 # SPDX-License-Identifier: AGPL-3.0
 
-import logging
-import ciso8601
-import arvados.util
-import re
-import csv
-import math
 import collections
+import csv
 import json
-from datetime import date, datetime, timedelta
-from typing import Dict, List
+import logging
+import math
+import re
 import statistics
 
 from dataclasses import dataclass
+from datetime import date, datetime, timedelta
+from typing import Dict, List
+
+import arvados.util
+import ciso8601
+
 from arvados_cluster_activity.prometheus import get_metric_usage, get_data_usage
 from arvados_cluster_activity.reportchart import ReportChart
-
 
 @dataclass
 class WorkflowRunSummary:
--- a/tools/cluster-activity/arvados_cluster_activity/report.py
+++ b/tools/cluster-activity/arvados_cluster_activity/report.py
@@ -247,7 +247,7 @@ class WorkflowRunSummary:
         self.cost += run.cumulative_cost()
         self.count += 1
         self.runtime += runtime
-        self.users.add(run.user['uuid'])
+        self.users.add(run.request['modified_by_user_uuid'])
 
     def total_cost(self):
         return self.cost
--- a/tools/cluster-activity/arvados_cluster_activity/report.py
+++ b/tools/cluster-activity/arvados_cluster_activity/report.py
@@ -5,185 +5,635 @@
 
 import collections
 import csv
+import dataclasses
+import datetime
+import enum
+import functools
+import itertools
 import json
 import logging
-import math
 import re
 import statistics
+import urllib.parse
 
-from dataclasses import dataclass
-from datetime import date, datetime, timedelta
-from typing import Dict, List
+from typing import Callable, ClassVar, Dict, List, Mapping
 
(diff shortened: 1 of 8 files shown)


tools/cluster-activity/arvados_cluster_activity/report.py
```diff
@@ -3,186 +3,637 @@
...
-    hours: float = 0
-    activityspan: str = ""
-    tablerow: str = ""
+@dataclasses.dataclass
+class BytesFormatter:
+    """Function to format a number of bytes as a human-friendly string"""
+    exp: int
+    suffixes: list[str]
 
+    def __call__(self, bytecount):
```
ZoeCMa, line 36: Although I've no doubt that the code is correct, could you add some simple unit tests for the small, formatting utilities such as this one, `duration_hms_fmt`, and the cost calculator, etc.?

tools/cluster-activity/arvados_cluster_activity/report.py
```diff
@@ -3,186 +3,637 @@
...
-def aws_monthly_cost(value):
-    value_gb = value / (1024*1024*1024)
-    first_50tb = min(1024*50, value_gb)
-    next_450tb = max(min(1024*450, value_gb-1024*50), 0)
-    over_500tb = max(value_gb-1024*500, 0)
 
-    monthly_cost = (first_50tb * 0.023) + (next_450tb * 0.022) + (over_500tb * 0.021)
-    return monthly_cost
+bytes_base2_fmt = BytesFormatter(1024, ['KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB'])
+format_with_suffix_base2 = bytes_base2_fmt
```
ZoeCMa: I'm not entirely sure why we do the aliasing here, rather than importing and using the likes of `bytes_base2_fmt` in `main.py`? Is this a part of a public API?

tools/cluster-activity/arvados_cluster_activity/report.py
```diff
@@ -241,90 +673,94 @@ def collect_container_hours(timestamp, value):
...
             """)
 
+        projects = sorted(
+            self.summaries.items(),
+            key=lambda kv: kv[1].total_cost(),
+            reverse=True,
+        )
+        project_rows = {
+            key: '<td>{0}</td> <td>{1}</td> <td>{2}</td> <td>{3}</td>'.format(
+                ', '.join(summary.usernames(self.users)),
```
ZoeCMa, line 708: Here IMO is _possibly_ a HTML interpolation slot that may be controlled by an untrustworthy source (the user, who may be able to put arbitrary things in their `full_name` [i.e. `first_name` and `last_name`]?). I think for extra safety, perhaps it won't hurt to `html.escape()` the value of slot 0.

I _think_ for the other uses of the pattern `"<some html>".format(args...)`, the interpolated data are not actually arbitrary; their format are numbers, dates, etc, so we likely don't have to prioritize them for now.

ZoeCMa (commented):
Hi @brettcs, thank you for the heroic effort with the rewrite :)

Please refer to the inline review comments. Thx.

tools/cluster-activity/arvados_cluster_activity/report.py
ZoeCMa: Just to be certain, I created a non-admin user using Google login on my rogue cluster and yes, they can change their own name, maybe to little bobbie tags.

tools/cluster-activity/arvados_cluster_activity/report.py
brettcs: Done in 780e54bdc6.

tools/cluster-activity/arvados_cluster_activity/report.py
brettcs: No, just because this was such a large branch already, I kept the old names just to maintain the internal API and reduce the number of files I needed to touch. Cleaned up in 60cc8d0449.

tools/cluster-activity/arvados_cluster_activity/report.py
brettcs: This is a preexisting bug. I would like to defer this, partly because I would like to address this by removing all the HTML generation code and rendering a Jinja template instead. Jinja already provides HTML escaping so we don't need to reinvent that particular wheel.

brettcs:
Updated branch tests at https://ci.arvados.org/job/developer-run-tests/5116/

tools/cluster-activity/arvados_cluster_activity/report.py
ZoeCMa: Thanks, I like the idea of doing proper HTML with jinja :)

ZoeCMa:
@brettcs thank you; LGTM.

ZoeCMa (approved):

Issue #456, opened 2026-07-16
Remove SDK CLI


brettcs:
Once the Python SDK installs the `arv` command, there is no longer any need for it.

brettcs:
Branch 456-remove-cli-sdk - <https://ci.arvados.org/job/developer-run-tests/5120/>

Ideally this would wait until after #455 is done. I wanted to get started because I thought it might be more involved but turns out it was easy enough.

brettcs:
Done in 2b6134a6a21410a2d881bfbabf50bfa880136039.

Issue #421, opened 2026-06-16
Migrate CLI docs to Python SDK


brettcs:
As part of retiring the CLI SDK, we're going to migrate all command line tool documentation under the Python SDK. I have pushed a branch 421-cli-doc-move-wip that shows how I think this should be structured. Please build on top of ef422a6235157174d6a478f347db689216925ccb. From there, "all" that remains is to actually write the page of command line tool documentation as described in the comments.

A lot of the subcommand descriptions and examples in the original CLI SDK are basically fine. If you would like to reuse those to write the new page, feel free. Just please note the following:

* All example UUIDs should follow our modern example format: `zzzzz-RTYPE-12345abcde67890`
* Do not include any `--help` output, as explained in the comment.

The new CLI documentation page should have a table of contents with links for easy navigation. Please follow the layout used by the Python code cookbook page.

ZoeCMa:
> Do not include any `--help` output, as explained in the comment.

Could you tell me what "the comment" means here? Which comment?

brettcs:
> Could you tell me what "the comment" means here? Which comment?

The comment in `doc/sdk/python/arv-cli.html.textile.liquid` (added by e56397a104564c3f8b58e0f3a651cb3a4c078fab) that outlines how the page should be written.

ZoeCMa:
Ah, it seems in e56397a104564c3f8b58e0f3a651cb3a4c078fab the file `doc/sdk/python/arv-cli.html.textile.liquid` was missing. Could you please git-add it? Thx.

brettcs:
Sorry for the oversight. Now at ef422a6235157174d6a478f347db689216925ccb.

ZoeCMa:
Thank you!

I wonder if there's a way to automate the render of the table of contents.

Perhaps out of the scope of this ticket, but may worth some further thoughts.

brettcs:
> I wonder if there's a way to automate the render of the table of contents.

I looked when I wrote the code cookbook and didn't find anything. If you want to search with fresh eyes for an existing solution in our documentation stack, be my guest. But please don't implement anything custom.

PR #459, opened 2026-07-17
455: Define scripts in Python SDK's `pyproject.toml` file

### LINKED ISSUE #455: Define project.scripts in sdk/python/pyproject.toml
In pyproject.toml, the `[project.scripts]` table lets you define executables for the package by pointing at functions. The build stack automatically generates a platform-approriate script to invoke that function. See contrib/arvados-bootstrap/pyproject.toml for an example.

Update the Python SDK to use this:

* Remove the `[tool.setuptools]` table.
* Add an equivalent `[project.scripts]` section.
* Add an entry for `arv` to `[project.scripts]`, since it's ready to ship.

ZoeCMa:
`455-define-scripts-in-pysdk-pyproject.toml` @ c8fb7aa0d98ee31a7a056e6bf8397abb7a8e298f

https://ci.arvados.org/job/developer-run-tests/5123/

  * This is based on `456-remove-cli-sdk` @ a46c6072f50fa2411d1d27a3bd161edff8a86112. (cf. #456)
  * The handcrafted executable scripts under `sdk/python/bin` are removed. `bin/arv-normalize` becomes the module `arvados.commands.normalize`, with a `main()` function.
  * `pyproject.toml` is updated to define scripts under section `[project.scripts]`
    * This includes the script `arv` which is defined to run `arvados.commands.arvcli`
    * `tests/test_arv_script_invocation.py` performs basic sanity check for all the new executable scripts. This test script is written to be PyTest-native.
    * Older tests now obsoleted by `tests/test_arv_script_invocation.py` are removed ("51 insertions(+), 105 deletions(-)" under `tests/` according to `git diff`)
  * _N/A_
  * Automated tests: see above.
  * Manual tests: installing the git head version by `pipx install .`
    * Observe the installation process to make sure all is well.
    * Inspect the generated `arv*` scripts in the venv.
    * Run the scripts to check that they behave without surprises.
    * Run `pipx uninstall arvados-python-client`.
  * _Yes_
  * _N/A_
  * See #421 and 5b5063e28a6d8a41fc5869d5779024a7a4652e39
  * For the scripts themselves, I'm not aware of major scaling differences between the Python CLI and the Ruby one w.r.t. size of the request data etc.
  * For deployment on client hosts, cutting off RubyGems dependency should be a win at scale.
  * If I understand correctly, for the Python CLI compatibility with the server hinges on the direct dependency of the upstream `googleapiclient` package, while for the Ruby CLI there's a level of indirection via our own fork of `ruby-google-api-client`. Compatibility should be easier to maintain for the former.
  * _Yes_ for new code.

About the deleted test `tests/test_arv_normalize.py`: all it did was sanity checks, and it has been entirely replaced by `tests/test_arv_script_invocation.py`

Closes #455.


ZoeCMa:
Also technically, the [entry point specification](https://packaging.python.org/en/latest/specifications/entry-points/#use-for-scripts) expects our `main()` functions to return an `int` status that's passed to `sys.exit()`. We aren't doing this, but we've made our `main()` functions `NoReturn` (#436, #451) (i.e. they do the exiting on their own) anyway.

PR #462, opened 2026-07-23
454: Improve cluster-activity argument parsing

### LINKED ISSUE #454: Improve cluster activity argument parsing
The argument parsing for the cluster activity report reimplements a lot of functionality that's already in argparse. It parses arguments and returns them as separate variables rather than using `type`. It logs errors and exits rather than using `argparser.error`.

Please clean all this up:

* Return and use everything under `args` with the right types. Write custom type functions as necessary to support this. Date options should return a timezone-aware datetime with the time set to midnight UTC. It should always use midnight UTC regardless of the user's timezone.
* Use `parser.error` to report higher-level argument errors.
* Let the user specify `--start` without `--end`: the help text already specifies that the default `--end` is today, and that is a fine default to use if the user only specifies `--start`. In other words, the only required user input is that the user specify _either_ `--start` _or_ `--days`. As you modernize the code, you can eliminate the `if (args.start and not args.end)` check.

ZoeCMa:
`454-cluster-activity-argument-parsing` @ 303d6152d0224a7304997c10877554aeeb2c2b14

https://ci.arvados.org/job/developer-run-tests/5130/

  * Custom types are used to convert and validate the CLI options `--start`|`--days`, `--end`, `--prometheus-auth`, and `--columns`
    * For the date arguments, the logic of mutual-exclusion, default, and conversion is implemented using the machinery of `argparse` and completed during the arg parsing.
    * Valid values for `--columns` are no longer hard-coded. They now automatically sync with the keys returned by `report.WorkflowRun.csv_row`.
  * _N/A_
  * Tests have been added and they're passing.
  * Manual test: inspecting the output of `arv-cluster-activity` usage and error messages to gauge the CLI UX.
  * _Yes_
  * _N/A_
  * _N/A_ (the changes are functionally compatible and doesn't require changing user instructions)
  * There _may_ be some penalty if a truly large number of CSV rows are to be written, because `report.WorkflowRun.csv_row` has become more complex, with the creation and destruction of the small, ephemeral dataclass instances (partly mitigated, for spatial efficiency, by the use of `@dataclass(slots=True)`). However, I suspect that the real speed bottleneck is the writing I/O itself, and that Python's object memory management should cope well with these small objects.
  * _N/A_
  * Yes for new code.

This PR also includes small fixes (descriptive names, etc.)

Closes #454.

### DIFF
--- a/tools/cluster-activity/arvados_cluster_activity/main.py
+++ b/tools/cluster-activity/arvados_cluster_activity/main.py
@@ -30,10 +30,10 @@ class _ArgTypes:
     """
     @staticmethod
     def date_arg(text: str | None = None) -> datetime:
-        """Convert input string in the form of YYYY-MM-DD to a UTC-timezone
-        datetime object for the specified date at 00h00m00s ("midnight").  If
-        the input is the empty string, return the midnight datetime of the
-        current UTC day.
+        """Converts the input string in the form of YYYY-MM-DD to a
+        UTC-timezone datetime object for the specified date at 00h00m00s
+        ("midnight"). When called without argument, returns the midnight
+        datetime of the current UTC day.
         """
         if text is None:
             dt_obj = datetime.now(timezone.utc)
@@ -54,7 +54,7 @@ class _ArgTypes:
 
     @staticmethod
     def positive_days(text: str) -> timedelta:
-        """Returns a timedelta of N days from the input string parsed as an
+        """Returns a timedelta of N days with the input string parsed as an
         integer. Negative values are invalid.
         """
         n = int(text)  # Let ValueError propagate.
@@ -64,7 +64,7 @@ class _ArgTypes:
 
     @staticmethod
     def n_days_before(src_arg: str) -> type[argparse.Action]:
-        """Make an `argparse.Action` subclass to be used by the "--days"
+        """Makes an `argparse.Action` subclass to be used by the "--days"
         argument. The returned subclass does the following:
             1. Take the value of the argument named `src_arg`;
             2. Subtract, from this `src_arg` value, the value of the argument
@@ -83,7 +83,7 @@ class _ArgTypes:
 
     @staticmethod
     def load_prometheus_auth(path: str) -> None:
-        """Load Prometheus environment variables from `path` into the current
+        """Loads Prometheus environment variables from `path` into the current
         process's `os.environ`.
         """
         prom_vars = {}
--- a/tools/cluster-activity/arvados_cluster_activity/main.py
+++ b/tools/cluster-activity/arvados_cluster_activity/main.py
@@ -208,26 +208,26 @@ def get_prometheus_client():
         logging.warn("Connecting to Prometheus failed, will not collect activity from Prometheus.  Error was: %s" % e)
         return None
 
-def report_from_prometheus(prom, cluster, since, to):
+def report_from_prometheus(prom, cluster, start_time, end_time):
 
     if not cluster:
         arv_client = arvados.api()
         cluster = arv_client.config()["ClusterID"]
 
-    print(cluster, "between", since, "and", to, "timespan", (to-since))
+    print(cluster, "between", start_time, "and", end_time, "timespan", (end_time - start_time))
 
     try:
-        print_data_usage(prom, since, cluster, "at start:")
+        print_data_usage(prom, start_time, cluster, "at start:")
     except:
         logging.exception("Failed to get start value")
 
     try:
-        print_data_usage(prom, to - timedelta(minutes=240), cluster, "current :")
+        print_data_usage(prom, end_time - timedelta(minutes=240), cluster, "current :")
     except:
         logging.exception("Failed to get end value")
 
-    print_container_usage(prom, since, to, "arvados_dispatchcloud_containers_running{cluster='%s'}" % cluster, '%.1f container hours', lambda x: x/60)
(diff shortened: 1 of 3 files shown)


ZoeCMa:
Hmmmm as I write this I think perhaps `typing.TypedDict` may be more suitable than a dataclass for "typing" the return value of `csv_row()`... but, `TypedDict` cannot enforce runtime checking of missing key etc.

tools/cluster-activity/arvados_cluster_activity/main.py
```diff
@@ -4,98 +4,199 @@
...
+        """Returns a timedelta of N days with the input string parsed as an
+        integer. Negative values are invalid.
+        """
+        n = int(text)  # Let ValueError propagate.
+        if n < 0:
+            raise ValueError(f"not a positive integer: {text!r}")
+        return timedelta(days=n)
+
+    @staticmethod
+    def n_days_before(src_arg: str) -> type[argparse.Action]:
```
brettcs: This implementation makes the argument parsing order-sensitive in a way that can become internally inconsistent and I think would surprise users even if it didn't. For example, running this version with `--days=90 --start-date=2025-12-31`, `args` becomes:

```python
Namespace(start=datetime.datetime(2026, 4, 24, 0, 0, tzinfo=datetime.timezone.utc), end=datetime.datetime(2025, 12, 31, 0, 0, tzinfo=datetime.timezone.utc), …)
```

IMO it would be simpler to implement, and give users the expected result, if you just postprocessed args after initial parsing:

```python
if args.start is None:
    args.start = args.end - args.days
```

As long as you do all the postprocessing _once_, _immediately_ after parsing, I don't see a problem with it. (It becomes more problematic when it gets scattered around the code.)

tools/cluster-activity/arvados_cluster_activity/report.py
```diff
@@ -844,12 +871,14 @@ def add_card(header, suffix):
...
-        else:
-            columns = ("Project", "Workflow", "Sample", "User", "Submitted", "Runtime", "CumulativeCost")
+        if not columns:
+            columns = ((
+                "Project", "Workflow", "Step",
+                "Sample", "User", "Submitted", "Runtime", "Cost"
+            ) if include_steps else (
+                "Project", "Workflow",
+                "Sample", "User", "Submitted", "Runtime", "CumulativeCost"
+            ))
```
brettcs: This is purely subjective so you can ignore this comment if you want, but my two cents is I find this version is harder to read than the previous version. It feels easy to miss the condition between two multi-line tuples.

brettcs:
TypedDict would be a nicer solution if we actually ran a type checker. Then the type checker could confirm that the argument parser and `csv_row()` stay in sync.

Since we don't run one, the solution you implemented is more real. Yes it does cost more at runtime, but that's the price we pay for doing this checking at runtime rather than as a type check at test time.

tools/cluster-activity/arvados_cluster_activity/report.py
```diff
@@ -98,6 +98,33 @@ def aws_monthly_cost(value):
...
+    User: str
+    UserUUID: str
+    Submitted: str
+    Started: str
+    Finished: str
+    Runtime: str
+    Cost: str
+    CumulativeCost: str
+
+    def asdict(self, **kwargs) -> dict[str, str]:
```
brettcs: This is also purely subjective, but I just don't see any value in having a method that literally just wraps a stdlib function.

tools/cluster-activity/arvados_cluster_activity/main.py
ZoeCMa: Not sure if I follow... In the arg parser builder, we have the following:

```python
    start_date_group = arg_parser.add_mutually_exclusive_group(required=True)
    start_date_group.add_argument(
        '--start',
        type=_ArgTypes.date_arg,
        help='Start date for the report in YYYY-MM-DD format (UTC)',
        metavar="DATE"
    )
    start_date_group.add_argument(
        '--days',
        type=_ArgTypes.positive_days,
        dest="start",
        action=_ArgTypes.n_days_before("end"),
        help='Number of days before "end" to start the report',
        metavar="N"
    )
```
The exclusive group means that `--start` and `--days` (two different ways to do the same thing -- writing to `args.start`) cannot coexist.

But, I believe you're right for a different reason -- if "do[ing] all the postprocessing _once_, _immediately_ after parsing" allows me to remove a lot of code, that's a plus. (I like removing code, especially that which I've written myself 😸)

ZoeCMa: Done in e5703a7814cab6ae960b6c6367a5a5b901d434b9.

tools/cluster-activity/arvados_cluster_activity/report.py
ZoeCMa: This is actually a good point. It was an artifact from my fussing around with `dataclasses`, and is fixed in 09eff886bcd5ef0477611da780eec6e3acc4508a.

tools/cluster-activity/arvados_cluster_activity/report.py
ZoeCMa: Sanity restored for your eyes in 09eff886bcd5ef0477611da780eec6e3acc4508a.

ZoeCMa:
Updated testing run: https://ci.arvados.org/job/developer-run-tests/5131/

tools/cluster-activity/arvados_cluster_activity/main.py
brettcs: I'm sorry, I made a typo writing the comment, `--start` was supposed to be `--end` (which you can see in the resulting Namespace, which I actually copy+pasted).

PR #465, opened 2026-07-24
464: Replace `PyYAML` with `ruamel.yaml` in run_test_server.py.

### LINKED ISSUE #464: Replace PyYAML with ruamel.yaml in the testing environment
The `sdk/python/tests/run_test_server.py` script depends on PyYAML (package name `yaml`).

We already use `ruamel.yaml`, rather than PyYAML, in the new Python-based CLI client, because the former is actively maintained. `arvados-cwl-runner` already pulls `ruamel.yaml` as a dependency too.

It's therefore natural to remove PyYAML from our testing environment in favor of `ruamel.yaml`.

ZoeCMa:
`464-replace-pyyaml` @ eb99d1079c790606f5266b92ab13b9aae9e3ae3e

https://ci.arvados.org/job/developer-run-tests/5132/

  * The replacement of PyYAML in sdk/python/tests/run_test_server.py is mostly straightforward.
  * By default, ruamel.yaml [no longer tolerates duplicate keys](https://yaml.dev/doc/ruamel.yaml/api/#Duplicate_keys) in YAML. This has helped us find some duplicated keys in the services/api test fixtures.
  * I don't _think_ we need to list `ruamel.yaml` in `build/requirements.tests.txt` anymore, because the Python SDK already pulls it in.
  * _N/A_
  * Tested with clearing the temp space and run `build/run-tests.sh` from scratch.
  * _Yes_
  * _N/A_
  * _N/A_
  * _N/A_
  * _N/A_
  * _Yes_, I think.

Closes #464.

### DIFF
--- a/build/requirements.tests.txt
+++ b/build/requirements.tests.txt
@@ -7,9 +7,8 @@
 ### Requirements for run-tests.sh itself
 -r requirements.build.txt
 
-# Required by sdk/python/tests/run_test_server.py
-# which is run directly by run-tests.sh
-PyYAML
+# YAML support required by sdk/python/tests/run_test_server.py is already
+# provided by the PySDK dependency ruamel.yaml
 
 # yq is used by run-tests.sh directly and controller tests
 yq ~= 3.4
--- a/services/api/test/fixtures/container_requests.yml
+++ b/services/api/test/fixtures/container_requests.yml
@@ -125,7 +125,6 @@ completed:
   runtime_constraints:
     vcpus: 1
     ram: 123
-  mounts: {}
 
 completed-older:
   uuid: zzzzz-xvhdp-cr4completedcr2
@@ -151,7 +150,6 @@ completed-older:
   runtime_constraints:
     vcpus: 1
     ram: 123
-  mounts: {}
 
 completed_diagnostics:
   name: CWL diagnostics hasher
@@ -433,7 +431,6 @@ requester:
   runtime_constraints:
     vcpus: 1
     ram: 123
-  mounts: {}
 
 cr_for_requester:
   uuid: zzzzz-xvhdp-cr4requestercnt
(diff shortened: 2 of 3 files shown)


build/requirements.tests.txt
```diff
@@ -7,9 +7,8 @@
 ### Requirements for run-tests.sh itself
 -r requirements.build.txt
 
-# Required by sdk/python/tests/run_test_server.py
-# which is run directly by run-tests.sh
-PyYAML
+# YAML support required by sdk/python/tests/run_test_server.py is already
+# provided by the PySDK dependency ruamel.yaml
```
brettcs, line 11: I don't think this comment makes much sense if you don't know what came before it. Python packages declaring their dependencies correctly is the normal situation. The fact that `run_test_server.py` used to be a weird exception is why a comment was required in the first place. IMO you should just remove this.

services/api/test/fixtures/container_requests.yml
```diff
@@ -125,7 +125,6 @@ completed:
   runtime_constraints:
     vcpus: 1
     ram: 123
-  mounts: {}
```
brettcs, line 128: Can you please explain a little why this change is required or appropriate? In general I would expect container request fixtures to include mounts, and it's not clear why changing the YAML library leads to removing them.

ZoeCMa: Here the `mounts` key is duplicate, and `ruamel.yaml` says this is invalid, according to the spec (see https://yaml.dev/doc/ruamel.yaml/api/#:~:text=Duplicate%20keys).

I suspect that the empty value could be just a left-over artifact or something. The value in line 122 above looked like the legit one.

brettcs: Gotcha, my bad for not looking at the larger context. That's totally fine.

Issue #426, opened 2026-06-18
Cluster activity report should report historical status as much as possible


brettcs:
If you run a cluster activity report with an end date in the past, the entire "Cluster status as of END_DATE" pane is not rendered. This is intentional: limits in the way we keep and store data mean we can't present perfect numbers. But we could present close-enough numbers with a caveat. Make the following changes:

* In ClusterSummary, when we count Arvados items, add a filter for `['created_at', '<', (report.to timestamp)]`. `_to_filter` basically has the code to do this but it's on the wrong class, so you'll need to move it or something.
* Apply the following diff to the report template. If you need or want to change the implementation code in here, that's fine, but this diff shows the interface change I want: where the information gets added and the specific wording are both important.
```diff
diff --git i/tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2 w/tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2
index 02e950915c..50f771ce6b 100644
--- i/tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2
+++ w/tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2
@@ -56,7 +56,6 @@ SPDX-License-Identifier: AGPL-3.0
 <h1>{{ title }}</h1>
 {% endcall %}
 
-{% if report.to.date() == today %}
 {% call card() %}
 <h2>Cluster status as of {{ report.to|date }}</h2>
 <table class='aggtable'><tbody>
@@ -85,9 +84,13 @@ SPDX-License-Identifier: AGPL-3.0
       <td>{{ cluster.monthly_storage_cost()|cost or unavail }}</td>
     </tr>
 </tbody></table>
+<p>This report was generated on {{ today|date }}.
+{% if (today - report.to.date()).days >= 3 %}
+The count of total users and projects will not include any that were deleted between {{ report.to|datespan(today) }}.
+{% endif %}
+</p>
 {{ price_footnote }}
 {% endcall %}
-{% endif %}
 
 {% call card() %}
 <h2>Activity and cost over the {{ (report.to - report.since).days }} day period {{ report_span }}</h2>
```

PR #469, opened 2026-07-24
Generate cluster activity report with Jinja

brettcs:
458-cluster-activity-jinja @ 0efc5e3d982089879aa7a453ce1012dcaa9823a8 - https://ci.arvados.org/job/developer-run-tests/5134/

a3a75eae47 has the main functional changes. aa8066340e is purely a whitespace change: view it with `git show -w aa8066340e` for an easy-to-read diff to confirm. 0efc5e3d98 then adds necessary scaffolding to the test runner.

  * Yes
  * N/A
  * See above. New tests added for new components.
  * Yes, branch includes the recent merge of #454.
  * N/A
  * N/A
  * N/A
  * N/A
  * Yes


tools/cluster-activity/arvados_cluster_activity/report.py
```diff
@@ -551,321 +666,50 @@ def iter_steps(self):
...
+
+        return jinja.get_template('report.html.j2').render(
+            active=active,
+            cluster=cluster,
+            cluster_id=arv_config['ClusterID'],
+            report=self,
+            report_span=datespan_fmt(self.since, self.to),
+            statistics=statistics,
+            summaries=summaries,
+            today=datetime.date.today(),
```
ZoeCMa, line 716: Because `datetime.date.today()` returns the current local date, this seems to go against our convention of using UTC dates for everything including user input on the CLI (and if I understand correctly, Prometheus already reports everything in UTC dates/times too). Is this the case?

tools/cluster-activity/arvados_cluster_activity/report.py
```diff
@@ -551,321 +666,50 @@ def iter_steps(self):
...
-            ('Storage usage', 'used'): storage_graph,
-        }
-        if any(graphs.values()):
-            cards.append("""
-                <div id="chart"></div>
-            """)
-
-        projects = sorted(
+        arv_config = self.arv_client.config()
+        active = self.summaries.pop(None)
```
ZoeCMa, line 677: This line is inherited from the base commit. It seems to cause an error if this method `html_report` is called twice in a row (not that we do that in our code!)

If you feel that this is a real issue, please feel free to change it. I may be wrong but I suspect this might indicate some underlying logic issue.

The following shim triggers this issue:

```python
import datetime
import arvados_cluster_activity.report
import arvados_cluster_activity.main

start = datetime.datetime(2026, 7, 25, 0, 0, 0, tzinfo=datetime.timezone.utc)
end = datetime.datetime(2026, 7, 28, 0, 0, 0, tzinfo=datetime.timezone.utc)
arvados_cluster_activity.main._ArgTypes.load_prometheus_auth(
    "/home/zcm/.config/prometheus-auth"
)
prom = arvados_cluster_activity.main.get_prometheus_client()
car = arvados_cluster_activity.report.ClusterActivityReport(
    start, end, prom_client=prom
)
h = car.html_report()
h = car.html_report()
```

```python-traceback
  File "/home/zcm/work-scraps/html-report/shim.py", line 16, in <module>
    h = car.html_report()
  File "/home/zcm/.cache/arvados-test/VENV3DIR/lib/python3.10/site-packages/arva
dos_cluster_activity/report.py", line 677, in html_report
    active = self.summaries.pop(None)
KeyError: None
```

tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2
```diff
@@ -0,0 +1,256 @@
...
+    {{ project_row(summary, key) }}
+    {% endfor %}
+  </tbody>
+</table>
+{{ price_footnote }}
+{% endcall %}
+
+{% for pkey, psum in summaries %}
+{% call card() %}
+<a id="{{ pkey }}"></a><h2><a href="{{ ('projects', psum.owner_uuid)|workbench_url }}">{{ pkey }}</a></h2>
```
ZoeCMa: To my surprise here, Jinja2 doesn't automatically replace whitespace characters in the `id` attribute value ([whitespaces are invalid](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/id#syntax) in this context). However, the string used as the `id` here (project name) is basically free text; it can and often does include the space character.

Perhaps something like this would eliminate the majority of whitespace problems:

```jinja2
<a id="{{ pkey |replace(' ', '-') }}">
```

(A related, potential issue is the uniqueness of `id`s. A project named "chart", for example, _may_ render the HTML invalid and potentially interfere with the Dygraph JS library. If you feel like doing deduplication on the Python side, please feel free to do so, or assign me to do it as another ticket; but personally I don't think it's really a priority for now.)

(Edited to remove distracting misinfo)

ZoeCMa (changes requested):
Thank you @brettcs 😃

My general comment is that I feel somewhat uncomfortable with the idea wrapping static resources files (CSS, JavaScript) in `{% raw %}` Jinja2 tags and letting them be templates.  For one thing, this tends to confuse text editors.  For another, it's a rather roundabout way to include files verbatim.

I suggest that we may add a function for raw-include:

```patch
diff --git a/tools/cluster-activity/arvados_cluster_activity/report.py b/tools/cluster-activity/arvados_cluster_activity/report.py
index 5652ad5695..fa0c9f53e4 100644
--- a/tools/cluster-activity/arvados_cluster_activity/report.py
+++ b/tools/cluster-activity/arvados_cluster_activity/report.py
@@ -692,12 +692,18 @@ class ClusterActivityReport:
             loader=jinja2.PackageLoader('arvados_cluster_activity'),
             trim_blocks=True,
         )
+
+        def include_file(path: str) -> markupsafe.Markup:
+            content, _, _ = jinja.loader.get_source(jinja, path)
+            return markupsafe.Markup(content)
+
         jinja.filters.update(_JINJA_FILTERS)
         wb_url = arv_config['Services']['Workbench2']['ExternalURL']
         def workbench_url(value):
             if not isinstance(value, str):
                 value = PurePath(*value).as_posix()
             return markupsafe.Markup(urllib.parse.urljoin(wb_url, value))
+        jinja.globals['include_file'] = include_file
         jinja.filters['workbench_url'] = workbench_url
 
         return jinja.get_template('report.html.j2').render(
```

And in the main template file, 

```patch
diff --git a/tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2 b/tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2
index 1e89286be8..02d61327c2 100644
--- a/tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2
+++ b/tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2
@@ -40,12 +40,14 @@ SPDX-License-Identifier: AGPL-3.0
           });
       });
 
-{% include 'synchronizer.js.j2' without context %}
-{% include 'dygraphs.js.j2' without context %}
-{% include 'sortable.js.j2' without context %}
+{{ include_file('synchronizer.js') }}
+{{ include_file('dygraphs.js') }}
+{{ include_file('sortable.js') }}
     </script>
 
-{% include 'report.css.j2' without context %}
+    <style>
+{{ include_file('report.css') }}
+    </style>
 
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/dygraph/2.0.0/dygraph.min.css">
   </head>
```

(I.e. the `.js.j2` and `.css.j2` files will become non-templates again, and the file `report.css` will also have its content unwrapped from the `<style>...</style>` tags and become pure CSS. The `MANIFEST.in` file would need to be updated too.)

Please let me know how you think about this idea.

tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2
ZoeCMa: Ah, silly me, I forgot to use `autoescape=jinja2.select_autoescape()` in this example when calling `jinja2.Environment()`. Sorry, please disregard the stuff about escaping. Perhaps `<a id="{{ pkey |replace(' ', '-') }}">` should suffice for now.

tools/cluster-activity/arvados_cluster_activity/report.py
brettcs: It sounds like we maybe weren't on the same page about the scope of the decision in that previous ticket. I thought we were just talking about how to turn the user command line argument strings into Python date/time objects. I didn't think we were setting a rule for all datetimes across the program.

In general, the program logic only cares about dates. When we talk to other APIs like Prometheus or Arvados, sometimes we need to specify a time, and it should always be midnight UTC when we do. So we settled on encoding the arguments that way because it fit better with the existing structure of the program.

But for internal program logic, working with dates is more straightforward, and I think you can see that if you look at where `today` gets used:

```jinja
{% if report.to.date() == today %}
```

`report.to` comes from `args.start`, so we're turning the argument into just a date and comparing against today. The other way to do this would be to create a datetime object representing today at midnight UTC, then compare against that. But I feel like adding that code just obscures the real purpose of what we're trying to do without any real benefit, because again, I don't think we need a global rule about how every single date/time is represented.

tools/cluster-activity/arvados_cluster_activity/report.py
brettcs: Yeah I consider this not an issue. We are not building a general-purpose library, we're just organizing the implementation of the report tool. It is okay to have "API" limitations that don't affect the tool.

To be clear, the _only_ way you can call more than one report method is to do what the code does now: `csv_report` then `html_report`. Doing anything else won't work because `iter_and_tally_runs` only actually iterates the first time you call it. This is an optimization because `html_report` doesn't actually want to iterate, just use the tallies, so it early-returns if it has already tallied something.

tools/cluster-activity/arvados_cluster_activity/templates/report.html.j2
brettcs: Fixed in 9fc46666bf by just using the UUID instead. This ensures a valid id without adding more code and it's guaranteed to be unique.

brettcs:
`include_file` added at feae8bf5c9. [developer-run-tests #5135](https://ci.arvados.org/job/developer-run-tests/5135/). Ready for another look.

tools/cluster-activity/arvados_cluster_activity/report.py
brettcs: I'll also add, I don't want to sweat this too much because I expect `today` to go away when we do #426.

ZoeCMa (approved):
I really like the idea of using UUIDs as `id` attribute values. Thanks.

This looks all good to merge!

PR #482, opened 2026-08-07
376: Support for upstream cwltool `INTERNAL_VERSION` v1.3.0-dev1

### LINKED ISSUE #376: arvados-cwl-runner needs to depend on a newer cwltool
The current dependency does not support Python 3.12+, so it blocks support for Debian 13 and Ubuntu 26.04.

This was originally blocked on https://github.com/common-workflow-language/cwltool/issues/2114. Trying to upgrade to a release with a fix still gets some test failures I haven't tracked down.

ZoeCMa:
`376-fix-support-for-cwl-internal-version` @ a01517a257f3a1d37c4a2a57f3d4295220330e8d

https://ci.arvados.org/job/developer-run-tests/5145/

  * When the CWL versions we support lag behind the so-called `INTERNAL_VERSION` of `cwltool`, re-use the latest CWL extension schema we support for that version.
  * This allows us to bump the dependency versions of `cwltool` and `schema_salad` in a-c-r, therefore clearing a blocker for supporting Python versions 3.13 or 3.14.
  * To be sure we're comparing the version strings correctly while not reinventing the wheel, we use `packaging.version.Version`, which pulls in the dependency on `packaging`.
  * Also sync the `ruamel.yaml` version pin in the Python SDK with the `cwltool` dependency.
  * _N/A_
  * This allows the sdk/cwl unit tests and integration tests to pass. For more (e.g. OOM tests), see discussion here https://github.com/arvados/arvados/issues/376#issuecomment-5197454808
  * _Yes_
  * _N/A_
  * _N/A_ (the change is largely under-the-hood; eventually we'll update the docs for Python versions supported)
  * _N/A_
  * Most likely not impacted.
  * _Yes_

Closes #376 .

### DIFF
--- a/sdk/cwl/arvados_cwl/__init__.py
+++ b/sdk/cwl/arvados_cwl/__init__.py
@@ -14,6 +14,8 @@ import os
 import sys
 import re
 
+from packaging.version import Version
+
 from schema_salad.sourceline import SourceLine
 import schema_salad.validate as validate
 import cwltool.main
@@ -22,6 +24,7 @@ import cwltool.process
 import cwltool.argparser
 from cwltool.errors import WorkflowException
 from cwltool.process import shortname, UnsupportedRequirement, use_custom_schema
+from cwltool.update import INTERNAL_VERSION
 from cwltool.utils import adjustFileObjs, adjustDirObjs, get_listing
 
 import arvados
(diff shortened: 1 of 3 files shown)


brettcs (approved):
This is perfect. Thank you very much for tracking this down, it's a huge help.

PR #502, opened 2026-08-28
496: Disable Files tab for trashed collection in WB.

### LINKED ISSUE #496: What to do with file listing of a trashed collection in Workbench?
Currently, in workbench, visiting a trashed collection from "Trash" will land us on the collection file listing tab, which simply displays a spinning wheel. The reason is that the WebDAV service, used by WB to retrieve listing and do other file operations, returns 404 Not Found when a trashed collection is requested.

The futile spinning seems... futile, and there are stuff we can do on the UI side to change that. For example, here's an idea I sketched out: disable the Files tab for a trashed collection.

https://github.com/user-attachments/assets/b1acac39-301e-4654-be6e-2fee113dd165

However, a related question is whether file listing should be available for trashed collections at all. I can see reasons for either "yes" or "no".

On the one hand, it seems reasonable to just say "a trashed collection is trashed", and if the file content is really needed, it's still possible to restore.  Also, since we'd likely have to implement read-only view of files if we're to enable it for a trashed collection, there's the question of why a user would want to do things with that read-only listing of file names, sizes, and directory structures.

The answer, on the other hand, may be that clues about file details in a trashed collection could come handy when the user needs to revert an erroneous trashing action, for collection names may not always be descriptive, and descriptions may be absent.

ZoeCMa:
`496-disable-files-tab-for-trashed-collection` @ 9c83332cb71cc4fb64807476c87ac638c4ced72b

https://ci.arvados.org/job/developer-run-tests/5156/

For an animation, see the post in #496.

  * For a trashed collection, the Files tab is disabled. It's still displayed, but greyed out and not clickable.
  * This is mostly to keep the visuals consistent with the existing version (and with normal collections), while removing the taste of contradiction from the UI of the existing version.
  * _N/A_
  * Automated (unit and integration) Cypress testing has been updated.
  * Manual testing: look and feel inspection, element inspection in browser devtools, manual trials with trashing/untrashing.
  * _Yes_
  * Well... this _is_ the place for feedback.
  * _N/A_
  * _N/A_ (one could argue that this reduces the futile `PROPFIND` requests associated with opening a trashed collection in Trash, but I don't think that's likely a scale issue.)
  * This doesn't require any change on the server side.
  * _Yes_

Closes #496.


services/workbench2/src/components/multi-panel-view/multi-panel-view.cy.js
```diff
@@ -97,6 +97,38 @@ describe('<MPVContainer />', () => {
         cy.contains('This is another panel');
     });
 
+    it('should configure disabled panels with tooltips', () => {
+        const childs = [
```
brettcs, line 101: Small nit but it would be nicer to name this `children`.

services/workbench2/src/components/multi-panel-view/multi-panel-view.tsx
```diff
@@ -121,6 +121,8 @@ function preventRerender(prevProps: MPVPanelContentProps, nextProps: MPVPanelCon
 export interface MPVPanelState {
     name: string;
     visible?: boolean;
+    disabled?: boolean;
+    disabledInfo?: string;
```
brettcs, line 125: I'm going to ask a question. I promise it's an honest question. I am asking because I think there might be an answer. I am not insinuating you *need* to make a change.

Is there some benefit to having two separate fields for this? It seems to me we could have just `disabledInfo?: string` by itself. Then we can calculate the `disabled` attribute with just `disabledInfo !== null`.

Put another way, are there situations where it makes sense, or is just convenient, to set `disabled: false` but also set `disabledInfo`?

brettcs: Separately, I think we can make the name more specific than `disabledInfo`. What about `disabledReason`, `disabledBecause`, or `disabledTooltip`?

services/workbench2/src/components/multi-panel-view/multi-panel-view.cy.js
ZoeCMa: Yeah this something I "inherited" from other tests in this spec file. I _think_ the intention was to avoid confusion with the destructured argument `children` of `PanelMock`. We can maybe rename every occurrence to `childPanelMocks`.

services/workbench2/src/components/multi-panel-view/multi-panel-view.tsx
ZoeCMa: It was just for convenience, and I think we can indeed name it `reasonToDisable` for both practical and expressive reasons. If the string value is empty or if the field is left `undefined`, it means "it has no reason to disable", hence it's not disabled. The boolean check for "disability" is then just `!!panelState?.reasonToDisable`. (Sorry I dislike the look of `!!` but I also dislike the jenga tower of booleans, truthiness, and the multiple nullities, so ....)

services/workbench2/src/components/multi-panel-view/multi-panel-view.cy.js
brettcs: Meh, like I said, it was a small nit. If it's following local convention, fine.

services/workbench2/src/components/multi-panel-view/multi-panel-view.tsx
brettcs: Sure, that's fine.

brettcs (approved):

PR #504, opened 2026-08-30
Add logrotate rule for RailsAPI server in Ansible

brettcs:
501-ansible-railsapi-logrotate @ 4a93ebd7e256df9ef0617d362bdee902fe428b5a - <https://ci.arvados.org/job/test-provision-ansible/40/>

This does what the Salt installer does except the default retention period is 30 days vs. Salt's 365. 30 is more suitable for smaller clusters which tend to do less configuration. Clusters migrating from Salt can use the new setting to retain their current behavior.

  * Yes
  * N/A
  * See above, also ran it on my local single-node cluster.
  * Yes
  * N/A
  * Yes (comment for the new setting as usual)
  * See notes above
  * Ditto
  * N/A

### DIFF
--- a/tools/ansible/roles/arvados_api/templates/logrotate.conf.j2
+++ b/tools/ansible/roles/arvados_api/templates/logrotate.conf.j2
@@ -1,0 +1,16 @@
+### This file is managed by Ansible ###
+{# Copyright (C) The Arvados Authors. All rights reserved.
+ #
+ # SPDX-License-Identifier: Apache-2.0
+ #}
+
+/var/www/arvados-api/shared/log/*.log
+{
+  {{ arvados_api_logrotate|indent(2) }}
+  missingok
+  copytruncate
+  sharedscripts
+  postrotate
+    systemctl try-reload-or-restart arvados-railsapi.service
+  endscript
+}
--- a/tools/ansible/roles/arvados_api/defaults/main.yml
+++ b/tools/ansible/roles/arvados_api/defaults/main.yml
@@ -2,5 +2,15 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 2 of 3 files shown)


tools/ansible/roles/arvados_api/templates/logrotate.conf.j2
```diff
@@ -0,0 +1,16 @@
...
+ #}
+
+/var/www/arvados-api/shared/log/*.log
+{
+  {{ arvados_api_logrotate|indent(2) }}
+  missingok
+  copytruncate
+  sharedscripts
+  postrotate
+    systemctl try-reload-or-restart arvados-railsapi.service
```
ZoeCMa, line 14: One quick question: what is the reload-or-restart for? I don't quite understand. Is it simply about replicating what the Salt installer used to do?

brettcs: This is necessary. At the highest level, this tells the running Passenger server process to reopen its log files. Without this, the server will keep writing to its already-open file descriptor and keep appending to the already-rotated log file. This situation is common in server daemons. If you look at all the files in your `/etc/logrotate.d` shipped with your distro, you'll probably find a few similar postrotate scripts.

The specific try-reload-or-restart subcommand means:

1. If the unit is not running, do nothing. (The server should not start just because we rotated logs.)
2. Otherwise, if the unit has a reload definition, run that. (Our unit file ships so the reload command is `bundle exec passenger reopen-logs` which does exactly what we need.)
3. Otherwise, restart the unit. (It shouldn't happen but maybe the administrator overwrote the reload command to remove it. In this case, restarting is a blunt instrument but makes sure log rotation is still effective.)

PR #505, opened 2026-08-31
503: Refactor "API Details" dialog's "API Response" tab for content/styling correctness & deduplicated code

### LINKED ISSUE #503: "API Response" tab in "API Details" dialog has wrong style, clips output
Currently in WB, the "API Response" tab from the "API Details" dialog looks like this:

Image

The `<pre>` content is meant to be a JSON document. The problem is that the opening and closing braces are somehow clipped.

Inspecting the element, we can see that currently the HTML structure looks like the following (slightly formatted and edited for clarity):

```html
<pre>
<span style="margin-left: -15px;">{ 
"uuid": "xsat1-j7d0g-a8p1lxzwbqrlh0e",
[...]
"can_manage": true 
<span style="margin-left: -15px;">}</span>
</span>
</pre>
```

These nested `<span>`s looks malformed. In fact, simply putting well-indented JSON string inside the `<pre>` should be fine.

Looking at `services/workbench2/src/store/advanced-tab/advanced-tab.tsx` which is responsible for the content generation here, there's a lot of duplicate code for the JSON-generation routines (where `JSON.stringify` should suffice). That should be fixed as well.

ZoeCMa:
`503-wb-api-response-content-styling-and-refactor` @ c2e29d16eb6583de20144a3b517d0fcb03ee0825

https://ci.arvados.org/job/developer-run-tests/5159/

  * Refactor `services/workbench2/src/store/advanced-tab/advanced-tab.tsx`:
    * No more manual generation of each JSON key-value pair; use standard `JSON.stringify()` for the whole object in one go.
    * DRY-up code.
    * Make the internal interface more declarative.
    * Remove unnecessary `<span>` element wrapping the displayed JSON.
  * Fix styling of displayed JSON output (opening/closing braces no longer swallowed).
  * Add tests to check that the JSON output is valid and correct.
  * Accessibility: No `ALL CAPS` in tab label.
  * _N/A_
  * Automated: new Cypress e2e tests for various resource types (collection, user, project).
  * Manual: inspecting the result in a browser.
  * _Yes_
  * _N/A_ (there's no major change in how the user perceives the interaction with WB, just the content)
  * _N/A_
  * _N/A_ (this change has no discernible effect on scaling)
  * The output displayed in WB is now actually valid JSON matching the server output, see below.
  * _Yes_

With the code before this change, the output may actually contain invalid JSON; e.g. for a user (slightly edited to remove the long value of `prefs.wb` key):

```json
{
"uuid": "xlogn-tpzed-dgrbpsawni869vr",
"owner_uuid": "xlogn-tpzed-000000000000000",
"created_at": "2026-07-20T07:07:55.704034000Z",
"modified_by_user_uuid": "xlogn-tpzed-dgrbpsawni869vr",
"modified_at": "2026-08-31T03:37:47.406551000Z",
"email": "zoe.ma@curii.com",
"first_name": "Zoë",
"last_name": ""Ma"",
"username": "zoema",
"is_active": "true,
"is_admin": "true,
"prefs": "{
  "profile": {},
  "wb": {}
},
"username": "zoema"
}
```

This was the drawback of crafting JSON by hand, which is fixed in this PR.

Also, in 8690d32798b4fcada70c2ec571cb8347f0003288 I fixed an unrelated test following the increase of snackbar timeout (from #490, 976ff21383226b23e085b328f4cd4e912ba4e6af) that only manifested as I worked on this branch. It's a bit amusing that this wasn't caught in previous test runs. This is basically an otherwise harmless race in the test setup, and changes to one part of the code might "act at a distance" and visibly affect another part, due to changes in load time, etc.

Closes #503.


services/workbench2/cypress/e2e/api-details-dialog.cy.js
```diff
@@ -0,0 +1,80 @@
...
+                .should((displayedData) => {
+                    expect(resource).to.deep.include(displayedData);
+                });
+        });
+}
+
+describe("'API Details' dialog showing valid 'API Response'", () => {
+    let adminUser;
+
+    before(function () {
```
brettcs, line 29: Am I following correctly that these tests only ever run as an admin user? I think they would be improved if they ran as a non-admin user. (Ideally they would run as both, but if we have to pick one for performance or practicality reasons, a non-admin user is a better choice.)

services/workbench2/cypress/e2e/api-details-dialog.cy.js
```diff
@@ -0,0 +1,80 @@
+// Copyright (C) The Arvados Authors. All rights reserved.
+//
+// SPDX-License-Identifier: AGPL-3.0
+
+function doTestWithResource(token, resourceType, testResource) {
```
brettcs, line 5: Suggest `testResourceAPIResponse` or `testResourceAPIDetails` as a clearer name.

services/workbench2/cypress/e2e/project.cy.js
```diff
@@ -621,7 +621,7 @@ describe("Project tests", function () {
         cy.contains("Project has been successfully created");
         cy.waitForDom();
         cy.get("[data-cy=form-dialog]").should("not.exist");
-        cy.get("[data-cy=snackbar]").should("not.exist");
+        cy.get("[data-cy=snackbar]", { timeout: 6000 }).should("not.exist");
```
brettcs, line 624: I have thoughts about this but I will make my comments on the separate ticket we have about test timeouts, we can address it there. This one change is fine for this branch.

services/workbench2/src/store/advanced-tab/advanced-tab.tsx
```diff
@@ -128,7 +129,13 @@ export const openAdvancedTabDialog = (uuid: string) =>
...
                     metadata: metaCollection,
                     user: userCollection,
-                    apiResponseKind: collectionApiResponse,
+                    apiResponseKeys: [
+                        'uuid', 'ownerUuid', 'createdAt', 'modifiedByUserUuid', 'modifiedAt', 'portableDataHash',
+                        'replicationDesired', 'replicationConfirmedAt', 'replicationConfirmed',
+                        'name', 'description', 'properties', 'deleteAt', 'trashAt', 'isTrashed',
+                        'storageClassesDesired', 'storageClassesConfirmed', 'storageClassesConfirmedAt',
+                        'currentVersionUuid', 'version', 'preserveVersion', 'fileCount', 'fileSizeTotal'
+                    ],
```
brettcs, line 138: I will just say, I do feel like it's a little sad how this code is basically re-hardcoding stuff that's already in the discovery document. It would be nice if we could make sure they are in sync.

However, what you've done here is still definitely an improvement over what we had before. I'm happy to take this branch as-is. I'm just mentioning it as a consideration for future development.

services/workbench2/cypress/e2e/project.cy.js
ZoeCMa: Please see #514 for further discussions.

services/workbench2/src/store/advanced-tab/advanced-tab.tsx
ZoeCMa: I admit I don't _fully_ grasp why it was done with these lists of properties the first time. My guess was that there was some kind of "sorting" and originally we (or someone) wanted the properties to appear in a certain order (maybe roughly grouped by semantics?)

But it is more likely that it's just a cognitive bias on my part (seeing pattern when there's none).

Is there any particular properties for certain resources that we do _not_ want to show in this "API Details" panel?

If there's none, personally I'd rather just display them in the original order.

brettcs: > Is there any particular properties for certain resources that we do not want to show in this "API Details" panel?

I don't see any that actually applies. I can imagine two possible reasons but neither of them currently apply to the code.

First reason would be security: it might be nice to avoid showing sensitive fields like `api_client_authorization.api_token` without the user specifically requesting it. But the current code shows it, so that's not it.

Second reason would be a usability concern: it might be nice to avoid showing arbitrarily large fields like `container_requests.mounts` or `collection.properties` that make it hard to see the rest of the response. But again, the current code doesn't hide any of those either.

So I'm fine with your plan of showing the original order.

services/workbench2/cypress/e2e/api-details-dialog.cy.js
ZoeCMa: Of course it works for the normal user. This is fixed in ce233d55333cb1ffac1dcda1ffc8ee9a85e839ff

services/workbench2/src/store/advanced-tab/advanced-tab.tsx
ZoeCMa: This is done in af0ec5411c34a82644a1ab1a9880091e3ab530cb.

services/workbench2/cypress/e2e/api-details-dialog.cy.js
```diff
@@ -0,0 +1,91 @@
...
+        .then((resource) => {
+            // In the content area, the <pre> element contains a valid JSON
+            // document that is a subset of the full API record.
+            cy.get(".MuiDialogContent-root pre")
+                .invoke("text")
+                .then(JSON.parse)
+                .then((displayedData) => {
+                    let filteredOutput = {};
+                    for (const [key, value] of Object.entries(displayedData)) {
+                        if (key !== "etag") filteredOutput[key] = value;
```
ZoeCMa, line 23: I had to do this because the `etag` for the collection is volatile. I had never managed to get a matching `etag` in the test. The `etag` is still displayed in the output on the page, just ignored in the test.

Issue #514, opened 2026-09-03
Brittle Cypress e2e tests


ZoeCMa:
EDIT: This post has been edited to expand the scope from one brittle test (the favorites SidePanel one) to possibly more

### `favorites.cy.js` SidePanel tests

This is the original topic and the most reliably triggered one.

A timeout happens with the custom command `cy.waitForDom` inside a `.then()` callback.

This begins to happen on Jenkins after updating to Cypress 14.x.

For an example, see https://ci.arvados.org/job/run-tests-services-workbench2-integration/1289

The test that triggers this issue is another one with a large, monolithic `.then()` callback chained after getting (querying) aliases.

### `project.cy.js` "copies project URL to clipboard"

See: https://github.com/arvados/arvados/pull/505#discussion_r3914944048

This is a snackbar timeout issue and has a simple fix.

### `collection.cy.js` file copy/move tests

See #491. (added as a sub-issue)

### `collection.cy.js` > "uploads and maintains nested folder structure"

Appears to be from this line:

https://github.com/arvados/arvados/blob/b4e3c0fc390ab95c9c9bf842361644a08704eee8/services/workbench2/cypress/e2e/collection.cy.js#L1310

The error message was:

```
     CypressError: Timed out retrying after 4050ms: `cy.click()` failed because 
the page updated while this command was executing. Cypress tried to locate eleme
nts based on this query:

> <div.Connect(Component)-rowName-182>

We initially found matching element(s), but while waiting for them to become act
ionable, they disappeared from the page. Common situations why this happens:
  - Your JS framework re-rendered asynchronously
  - Your app code reacted to an event firing and removed the element

You can typically solve this by breaking up a chain. For example, rewrite:

> `cy.get('button').click().click()`

to

> `cy.get('button').as('btn').click()`
> `cy.get('@btn').click()`

https://on.cypress.io/element-has-detached-from-dom
```

### `multiselect-toolbar.cy.js` "For users" > "should behave correctly for a single user"

https://github.com/arvados/arvados/blob/b4e3c0fc390ab95c9c9bf842361644a08704eee8/services/workbench2/cypress/e2e/multiselect-toolbar.cy.js#L1164

The intention here was to assert the snackbar, and it also runs into the timing issue. A rare error, only caught in live e2e testing in Cypress app.

### `banner-tooltip.cy.js` > "should re-show the banner"

[...]

PR #516, opened 2026-09-08
514: WB: fix favorites e2e tests

ZoeCMa:
`514-wb-fix-favorites-e2e-tests` @ 2dc7d19af1ff33f4815a339840103e8820e3870a

* For this branch, favorites e2e tests only:
  https://ci.arvados.org/job/developer-run-tests/5176/
* For recent WB testing changes (merged bundle of this branch, `491-clean-up-cypress-collection-e2e-tests-for-maintainability-and-reliability`, and `514-wb-various-e2e-test-fixes`)
  https://ci.arvados.org/job/developer-run-tests/5174/

  * The `favorites.cy.js` failures ([example](https://ci.arvados.org/job/run-tests-services-workbench2-integration/1289)) originated from the custom Cypress command `cy.waitForDom()` timing out. The error output is not very helpful, being a Cypress limitation when the command fails in a big monolithic `.then()` block.
  * To fix this, the offending test ("Favorites-SidePanel tests" > "shows the correct favorites and public favorites in the side panel") has been broken into three tests for better isolation:
    1. the original one "it shows the correct favorites and public favorites in the side panel" now only tests what it says in the title, that the correct items are shown / removed in the "My Favorites" and "Public Favorites" side panel sections.
    2. new test split off the original large test, "it restores trashed favorite project to favorites", to test restoration to "My Favorites" side panel of a faved but trashed project.
    3. another split-off test "it restores trashed favorite collection to favorites", similar to the above, for faved but trashed collection.
  * In these tests, we reduced the use of `cy.waitForDom()` which is resource-heavy (observing mutations on `window.document.body` and its entire subtree) and often introduces unpredictability.
  * When checking for the desired outcome in these tests, we used to do a `cy.wait(1000)` (unconditional pause of 1s) followed by a subtree text-content assertion on the container element (The "My Favorites" section in the side panel). The latter assertion may still timeout during the fairly XHR-heavy "restore resource with fave status and update UI" operation. In this fix, we instead get rid of the `cy.wait()` and use a different query-assertion strategy. We simply locate the parent (container) element, and query-and-assert precisely (using `data-id` attribute) on the expected child element that should contain the project/collection about to be restored, with a longer-than-default timeout.
  * We also applied some techniques for faster and reliable testing, such as setting up state by Arvados API calls rather than by Web UI (point and click).
  * _N/A_
  * See above
  * Repeat manual testing in Cypress app.
  * _Yes_
  * _N/A_ (no change)
  * _N/A_
  * _N/A_ (for test infrastructure only)
  * Not affected
  * _Yes_, mostly

Partially addresses #514.

### DIFF
--- a/services/workbench2/cypress/e2e/favorites.cy.js
+++ b/services/workbench2/cypress/e2e/favorites.cy.js
@@ -353,6 +353,15 @@ describe('Favorites-SidePanel tests', function () {
     });
 
     it('restores trashed favorite project to favorites', () => {
+        // The following faved item is not touched; simply used to keep the
+        // favorites dropdown in the sidebar from collapsing once (which
+        // happens if all faved items are gone)
+        cy.createProject({
+            owningUser: adminUser,
+            projectName: `myFavoriteProject0`,
+            addToFavorites: true,
+        });
+
         cy.createProject({
             owningUser: adminUser,
             projectName: `myFavoriteProject1`,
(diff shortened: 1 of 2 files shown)


services/workbench2/cypress/e2e/favorites.cy.js
```diff
@@ -358,55 +347,108 @@ describe('Favorites-SidePanel tests', function () {
...
+            });
+
+        cy.get('@myFavoriteProject1')
+            .then((myFavoriteProject1) => {
                 // Check project restored to favorites
-                cy.wait(1000);
-                cy.get('[data-cy=tree-item-toggle-my-favorites]').parents('[data-cy=tree-top-level-item]').should('contain', myFavoriteProject1.name);
-        });
+                cy.get('[data-cy=tree-item-toggle-my-favorites]').parents('[data-cy=tree-top-level-item]').within(() => {
+                    cy.get(`[data-id="${myFavoriteProject1.uuid}"]`, { timeout: 6000 }).should("contain", myFavoriteProject1.name);
```
brettcs: Where does the timeout value 6000 come from here? As best I can tell, nothing in this test is waiting for a snackbar to dismiss or anything, so I don't understand how we arrived at this value.

services/workbench2/cypress/e2e/favorites.cy.js
```diff
@@ -358,55 +347,108 @@ describe('Favorites-SidePanel tests', function () {
...
+        cy.get('[data-cy=side-panel-tree]').contains('Trash').click();
+        // Collection might not be on first page
+        cy.get('[data-cy=search-input] input').type(`${collName}{enter}`);
+        cy.get('[data-cy=data-table]').contains(collName).rightclick();
+        cy.get('[data-cy=context-menu]').contains('Restore').click();
+        // Check collection restored to favorites
+        cy.get("@testCollectionUuid")
+            .then((uuid) => {
+                cy.get('[data-cy=tree-item-toggle-my-favorites]').parents('[data-cy=tree-top-level-item]').within(() => {
+                    cy.get(`[data-id="${uuid}"]`, { timeout: 6000 }).should("contain", collName);
```
brettcs: Same question here.

services/workbench2/cypress/e2e/favorites.cy.js
ZoeCMa: Honestly this is just "150% times the Cypress default (4000)". The side panel takes rather long to sync when a faved project is untrashed.

I didn't feel like using `waitForDom()` because at the time of untrashing the UI is rather noisy and many elements come and go. I'm not sure the coarse-grained `waitForDom()` would be waiting for the desired state.

If you feel this is fine, I can add comments about the reason for this timeout.

brettcs: Yes please. I would also consider doubling the timeout to 8000. I'm sure 6000 is consistently passing for you, but the tests won't always be running on your hardware, and having them fail just because they didn't wait long enough is always rough. It's just a timeout, so it's not like it will wait the full eight seconds every time.

ZoeCMa: There's a test running after I pushed 1cb72c7ee2cbcf82065c0656eb67114ebb89ac60 for this: https://ci.arvados.org/job/developer-run-tests/5184/

PR #517, opened 2026-09-08
514: WB: various e2e test fixes

ZoeCMa:
`514-wb-various-e2e-test-fixes` @ 7d088f2b2bd217db3e155c674bfdcfedeca35dfc

* For this branch only:
  https://ci.arvados.org/job/developer-run-tests/5178/

* For recent WB testing changes (merged bundle of this branch, `491-clean-up-cypress-collection-e2e-tests-for-maintainability-and-reliability`, and `514-wb-fix-favorites-e2e-tests`)
https://ci.arvados.org/job/developer-run-tests/5174/

  * For the `banner-tooltip.cy.js` test, the issue is that Cypress clicks on a button too early (before the event listener has been attached), therefore failing to achieve the desired effect (i.e., to dismiss the modal dialog). The `cy.waitForDom()` custom command or the `timeout` option in Cypress queries are largely ineffective, because the expected change (event listener attaching) is virtually invisible to Cypress (hence also to its retry mechanism). There are possible [lower-level fixes](https://glebbahmutov.com/blog/when-can-the-test-click/) but it's way more complicated than is worth. Alternatively we could ["rage-click" the button](https://www.cypress.io/blog/when-can-the-test-click#:~:text=We%20can%20retry%20arbitrary%20user%20actions%20by%20using%20an%20ingenious%203rd%20party%20Cypress%20plugin%20called%20cypress%2Dpipe%2E) until it works, but in live testing I found the `cypress-pipe` solution creates DOM artifacts that made things worse. In the end I chose the [`cy.wait(500)` solution](https://www.cypress.io/blog/when-can-the-test-click#:~:text=%2Ewait%28500%29) that worked for both live and headless testing.
  * For the `multiselect-toolbar.cy.js` and `project.cy.js` tests, it's the same issue (proper technique for asserting the animation of snackbar). Here we use a longer-than-default timeout for the `cy.get()` query, but for slightly different reasons
    * For `multiselect-toolbar.cy.js` it's because the snackbar feedback of the UI operation is slow in itself (many things are happening in the UI at the same time), so we just need more time.
    * For `project.cy.js` it's about the disappearance of a particular snackbar. This needs to be at least the lifetime of the snackbar.
  * _N/A_
  * See above;
  * Manually tested locally, both headlessly and live.
  * _Yes_
  * _N/A_ (no change to user-facing code)
  * _N/A_
  * _N/A_ (not affected)
  * _N/A_
  * _Yes_ for new code.

Partially addresses #514.


services/workbench2/cypress/e2e/multiselect-toolbar.cy.js
```diff
@@ -940,10 +940,8 @@ describe('For workflow resources', () => {
 
             //delete workflow
             cy.doToolbarAction("Delete Workflow");
-            cy.get('[data-cy=confirmation-dialog]').within(() => {
-                cy.get('[data-cy=confirmation-dialog-ok-btn]').click();
-            });
-            cy.contains('Removed').should('be.visible');
+            cy.get('[data-cy=confirmation-dialog] [data-cy=confirmation-dialog-ok-btn]').click();
+            cy.get("[data-cy=snackbar]", { timeout: 10000 }).should("contain", 'Removed');
```
brettcs: I have the same question here as the other branch. Where does the 10000 value come from? Because we're waiting for snackbar *appearance*, not *dismissal*, I don't see a reason to tie this to the snackbar's own timeout. What am I missing?

brettcs: Same comment for the other two hunks later in this file.

ZoeCMa: If I remember this correctly the issue is that the rapid-burst automated action in Cypress creates not just one snackbar but many. There's a blue (info) "Removing ..." bar that won't be replaced by the green "Removed" in-place, but gets pushed up the stack of snackbars by "Removed". I need to double check this.

brettcs: That's fine. As in the other branch, I think it's worth adding an explanatory comment for this.

ZoeCMa: I found a more reliable way to do that without a long timeout: e63bad598b71fcc4c24d0cfad8e172b9da49d319

The rationale is to avoid querying a volatile container element for text changes in one of its child nodes, but to select the child nodes directly. Chaining the assertions to the container element (`[data-cy="snackbar"]`) would be fine if it doesn't change too quickly, but would be inefficient and possibly flaky if it does.

ZoeCMa: There's a test running: https://ci.arvados.org/job/developer-run-tests/5183/ (with an unrelated failure to start the Python SDK tests at all)

Issue #526, opened 2026-09-14
Update webpack and related dependencies for Workbench


ZoeCMa:
We're currently using Webpack 5.105.2 (released in February). For developing and testing, we use `webpack-dev-server` 4.15.2 (March 2024).

With the current Webpack version we use, we're seeing errors in Workbench unit (component) tests in Cypress like the following https://ci.arvados.org/job/run-tests-remainder/6099/

I'm also seeing the same error intermittently locally, for a different test. I suspect that upgrading to a more recent Webpack may be the first thing we can try to fix the flaky test.

`webpack-dev-server` is not just used for live development but also for running the tests. It's generally a good idea to keep it up-to-date.

ZoeCMa:
I already have a WIP branch for these version updates and I'll get around to that soon.

PR #527, opened 2026-09-14
514: harden clipboard tests in WB e2e test suite

### LINKED ISSUE #514: Brittle Cypress e2e tests

ZoeCMa:
`514-clipboard-in-e2e-tests` @ bc10f70bf2a4a77f3ba6a766f3f1a69f361acc98

https://ci.arvados.org/job/developer-run-tests/5187/ (❌ failed with most-likely-unrelated flake in WB unit tests)
https://ci.arvados.org/job/developer-run-tests/5188/ (✅ re-run with success)

  * Instead of doing long-timeout query on `cy.window()`, follow the Cypress example on clipboard testing: https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/testing-dom__clipboard/cypress/e2e/permissions-spec.cy.js
    * This also eliminates thenable-pyramiding as we switch to a chaining style, and makes the code easier to read.
  * A few URL regexes have been hardened (anchored to beginning and end, optional port part restricted to 5 digits) and have unnecessary escapes removed.
  * See #526.
  * See above for discussion about unrelated test that appeared flaky.
  * For local testing, I ran the changed e2e tests headlessly in a loop over 50 times and didn't encounter a failure.
  * _Yes_
  * _N/A_
  * _N/A_
  * _N/A_
  * _N/A_
  * _Yes_ for new code.

For more context, see https://github.com/arvados/arvados/issues/514#issuecomment-5658614227

Notice that I removed a `console.log` call that described itself in these words: `this [is] a load-bearing console.log`. I am really not sure how this came to pass. This was introduced in 93bb55716f95143cc42b9af328bc1aca1bc2ea2a (see old [redmine issue 22564](https://dev.arvados.org/issues/22564) where there was [a reference to the clipboard](https://dev.arvados.org/issues/22564#:~:text=There%20were%20several%20issues%20trying%20to%20test%20actions%20that%20copy%20to%20the%20clipboard%2E) without much detail).

I am willing to do the woodpecker thing and remove it, as it seems that the need to do what we did is no longer there.

References [`xkcd#2347`](https://xkcd.com/2347/ "xkcd comic episode 'Dependency'").
Closes #514.


ZoeCMa:
It's pretty weird. If I run the **old** clipboard e2e tests live (headed) in Cypress 15 as part of the whole spec, they "pass" as _false negatives_: the assertion fails but the test somehow "passes".

If I use `it.only()` to pick the (old) clipboard test and run that alone in the spec, the test fails. But, the copy-to-clipboard button works in the end-state after the test has failed.

In headless mode with Cypress 15, the whole spec fails at the clipboard test:

```
  1) For project resources
       "after each" hook for "handles project Copy UUID":

      AssertionError: The following error originated from your test code, not from Cypress. It was caused by an unhandled promise rejection.

  > expected '' to equal 'zzzzz-j7d0g-vqu4h1zejc9e8a7'

When Cypress detects uncaught errors originating from your test code it will automatically fail the current test.

Because this error occurred during a `after each` hook we are skipping all of the remaining tests.
      + expected - actual

      +'\'zzzzz-j7d0g-vqu4h1zejc9e8a7\''

      at eval (webpack://arvados-workbench-2/./cypress/e2e/multiselect-toolbar.cy.js:382:36)
```

Need to double check if the old tests have "false negatives" with the Cypress version (14) we currently use.

Under the hood, we are using some pretty old pattern for writing to the clipboard ([`document.execCommand("copy")`](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) anyone) from a 3rd party library and I wonder if that is causing trouble in Cypress.

ZoeCMa:
(I believe this was closed inadvertently when a different branch was merged).

ZoeCMa:
In addition, I do have a WIP that yanks out the underlying code that calls `document.execCommand()`, which seems to have largely resolved the issue. I'll write it up and let that supersede this.

PR #533, opened 2026-09-21
532: Fix labels to the copy-to-clipboard buttons.

### LINKED ISSUE #532: Workbench: "Copy link to clipboard" in labels or tooltips for clickables that doesn't copy "link"
These can be seen in two places:

1. In the details panel, the small "copy" icons have tooltips that says "Copy link to clipboard".
2. In the Overview tab of a resource's page.
3. In the "Get API Token" dialog under Account Management. In addition to the icons, there's a button with label "Copy link to clipBOARD" [sic] that actually copies the code block.

Each of these instances actually copies a text value (UUID, PDH, etc.) and not a link.

This should be fixed for consistency and also accessibility. The wrong tooltips on the small icons makes its `aria-label` accessible feature much less useful, especially as the UUID and PDH are formatted as clickable links (`<a>` tags) next to them. A user may be reasonably led to believe that the action was about copying the URL rather than the string value.

ZoeCMa:
`532-wb-fix-copy-button-label` @ b378ae827597273471536ccfd341ef35f0c7a766

https://ci.arvados.org/job/developer-run-tests/5199/

  * Small copy icon next to values in the details view: "Copy link to clipboard" -> "Copy value to clipboard"
  * "Command" tab in Process view: "Copy command to clipboard". Similarly for Logs tab.
  * "Get API token" page: "Copy link to clipBOARD" -> "Copy code to clipboard" (this is about copying the environment variable exporting code for shell).
  * _N/A_
  * Manual testing: in-browser inspections.
  * _Yes_
  * _N/A_ (it's a tiny change)
  * _N/A_
  * _N/A_
  * _N/A_
  * _Yes_

Closes #532.


ZoeCMa:
Guess what, there's another unrelated flaky e2e test.

